PR #142, opened 2020-01-30
Us112725/style the transcript

jwong-d2l:
This is a screenshot from the transcript page with print css enabled: 

image

This is a screenshot from the printed version (using the demo): 

image

Note that the `d2l-html-editor` isn't working properly in the demo page and that's why you only see a line instead of the answer. I tried to debug the issue but it was taking too long and I wanted to merge this rather quickly to unblock other work itens. However, my setup and use of the component seems to be correct given that it's working on the other page.


demo/report.html
```diff
@@ -1,10 +1,20 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en-gb" style="font-size:20px; font-family: 'Lato', sans-serif;">
```
TomMichielsD2L: Will these be given to use in bsi land?

demo/report.html
```diff
@@ -1,10 +1,20 @@
...
 	<head>
 		<script src="../node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
 		<script type="module">
 			import '../src/components/cpd-record-report';
 		</script>
+		<script type="module" src="../../core/components/typography/typography.js"></script>
```
TomMichielsD2L: Do we also get these in BSI land?

src/components/cpd-record-report.js
```diff
@@ -36,7 +41,67 @@ class CpdRecordReport extends BaseMixin(LitElement) {
...
+			}
+			.qa-text {
+				grid-column: span 2;
+				margin: 20px;
+			}
+			d2l-html-editor {
+				border-radius: 0.3rem;
+				border-style: solid;
+				border-width: 1px;
+				border-color: initial;
```
TomMichielsD2L: I thought we did not want a border on the html editor?

src/components/cpd-record-report.js
```diff
@@ -36,7 +41,67 @@ class CpdRecordReport extends BaseMixin(LitElement) {
...
+				margin-top: 60px;
+				margin-bottom: 20px;
+			}
+			p.attachment {
+				margin-top: 0.2rem;
+			}
+			@media print {
+				.container {
+					padding: 0px;
+					width: 7.5in;
```
TomMichielsD2L: Nice that you got print working with only a little bit of css

demo/report.html
jwong-d2l: Font sizes and families are provided by the blank page renderer.

demo/report.html
jwong-d2l: yep!

src/components/cpd-record-report.js
benyamin-noori: You're correct. I also reduced the top/bottom margins to make it look more like the sample transcript.

src/components/progress-subject.js
```diff
@@ -35,6 +35,9 @@ class ProgressSubject extends BaseMixin(LitElement) {
 			.progress-label {
 				vertical-align: middle;
 			}
+			d2l-meter-linear {
+				-webkit-print-color-adjust: exact;
```
jwong-d2l: this is to fix safari?

benyamin-noori: Chrome, actually!

src/components/cpd-record-report.js
```diff
@@ -80,47 +147,99 @@ class CpdRecordReport extends BaseMixin(LitElement) {
 			});
 	}
 
+	getQuestionKey(record, questionId) {
+		const answer = record && record.Answers && record.Answers.find(a => a.QuestionId === questionId) || false;
+		return answer ? 're-render' : 'render';
```
BraydonKains: Ahh I forgot about Tom's favourite HTML editor workaround 🤣

PR #150, opened 2020-02-04
Add time helper tests

jwong-d2l:


test/time-helper_tests.js
```diff
@@ -47,4 +47,31 @@ describe('TimeHelpers', () => {
...
+			Date.now = originalDateNow;
+		});
+		const mockDateFeb29 = () => {
+			return new Date(2020, 1, 29);
+		};
+		const mockDateNotFeb29 = () => {
+			return new Date(2020, 6, 12);
+		};
+		it('should return Feb 28th date if the current date is Feb 29th in a leap year', () => {
+			Date.now = mockDateFeb29;
```
jwong-d2l: We could use mockdate, but this seems fine to me.

src/helpers/time-helper.js
```diff
@@ -61,7 +61,7 @@ export function getNonLeapYearDate(month, date) {
 }
 
 export function getCurrentDate() {
-	const today = new Date();
+	const today = new Date(Date.now());
```
TomMichielsD2L: Why did we change this, for as far as I know, this is the same thing

PR #155, opened 2020-02-05
fix alignment on print

jwong-d2l:
image


demo/report.html
```diff
@@ -4,6 +4,7 @@
 		<script src="../node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
 		<script type="module">
 			import '../src/components/cpd-record-report';
+			import '../../../d2l-html-editor/demo/d2l-html-editor-client.js';
```
benyamin-noori: Did this fix the issue?

jwong-d2l: no, but it made the errors go away.

PR #156, opened 2020-02-05
DE37753 disable save button when clicked

jwong-d2l:


src/components/cpd-add-record.js
```diff
@@ -327,7 +330,7 @@ class AddCpdRecord extends BaseMixin(LitElement) {
 					${this.questions.map((q) => this.renderQuestion(q))}
 				</ul>
 				<div>
-					<d2l-button @click="${this.saveForm}">${this.localize('save')}</d2l-button>
+					<d2l-button ?disabled="${this.saving}" @click="${this.saveForm}" >${this.localize('save')}</d2l-button>
 					<d2l-button @click="${this.cancelForm}">${this.localize('cancel')}</d2l-button>
```
TomMichielsD2L: Should we also disable the cancel button?

jwong-d2l: we should let them click it.

PR #158, opened 2020-02-11
BSI Based Changes

BraydonKains:
https://rally1.rallydev.com/#/15545269080d/detail/task/367021260940?fdp=true
These are the changes needed for BSI compatibility. Ignore the branch name as it is no longer reflective of the changes. 😄


src/services/cpd-service-factory.js
```diff
@@ -1,11 +1,14 @@
...
 export class CpdServiceFactory {
 	static getCpdService() {
 		if (window.data && window.data.fraSettings && window.data.fraSettings.valenceHost) {
-			return CpdService;
+			return new CpdService(false);
+		} else if (window.demo) {
+			return DemoCpdService;
 		}
-		return DemoCpdService;
+		return new CpdService(true);
```
jwong-d2l: Would inheritance be a better way to share logic in this guy?

BraydonKains: I was having a hard time figuring out the right way to use inheritance for this... Maybe we can whiteboard it offline if you have a better idea.

jwong-d2l: the only thing we need to change are the options and the host, right?

BraydonKains: Yeah, the rest is the same I think.

jwong-d2l: https://stackoverflow.com/questions/5441508/how-to-inherit-static-methods-from-base-class-in-javascript

jwong-d2l: I dislike changing the way the service works just to support the changes to those two things.

src/components/cpd-record-report.js
```diff
@@ -218,14 +219,9 @@ class CpdRecordReport extends BaseMixin(LitElement) {
 				<div class="d2l-heading-3">
 					${question}
 				</div>
-				<d2l-html-editor
-					key="${this.getQuestionKey(this.records, question.Id)}"
```
TomMichielsD2L: We no longer need the getQuestionKey function on line 167

src/services/cpd-service-factory.js
```diff
@@ -1,11 +1,14 @@
+import { BsiService } from '../services/bsi-service';
 import { CpdService } from '../services/cpd-service';
 import { DemoCpdService } from '../services/demo-cpd-service';
 
 export class CpdServiceFactory {
 	static getCpdService() {
 		if (window.data && window.data.fraSettings && window.data.fraSettings.valenceHost) {
 			return CpdService;
+		} else if (window.demo) {
```
TomMichielsD2L: Did we already set this demo property on the window somewhere?

BraydonKains: Nope, this will break the demo for the time being.

src/components/cpd-record-report.js
BraydonKains: Will get this when I fix linting.

src/services/cpd-service-factory.js
BraydonKains: Update: In the fixes to the tests I opted to also fix the demos.

src/services/bsi-service.js
```diff
@@ -0,0 +1,18 @@
+import { CpdService } from './cpd-service';
```
jwong-d2l: so much cleaner! =)

BraydonKains: It took Tom and I spitballing on it for me to remember overriding methods was a thing. 😄

PR #160, opened 2020-02-13
Bump the package version so it's in sync with our current release ver…

BraydonKains:
…sion number

It was still 1.1.2 but our latest release was 1.1.3. It tried to bump to 1.1.3 and realized that it exists. Now when I trigger release off this, it will bump to 1.1.4 and put us back in sync.


BraydonKains:
This was the wrong way to do it

PR #162, opened 2020-02-13
Us112471/question

jwong-d2l:


package.json
```diff
@@ -17,7 +17,8 @@
     },
     "frauLocalAppResolver": {
       "appClass": "urn:d2l:fra:class:continuous-professional-development",
-      "dist": "./build/dev"
+      "dist": "./build/dev",
+      "hostname": "localhost"
```
jwong-d2l: this is fine because it's only for local

polymer.json
```diff
@@ -10,6 +10,10 @@
     "package.json",
     "index.html"
   ],
+  "lint": {
+    "rules": [],
+    "ignoreWarnings": ["non-literal-import"]
```
jwong-d2l: this ignores ben's thing

src/services/cpd-service.js
```diff
@@ -12,21 +12,68 @@ d2lfetch.use({
 });
 
 export class CpdService {
+
 	static CpdPath(action) { return `/d2l/api/customization/cpd/1.0/${action}`; }
+
+	static get Create() {
```
jwong-d2l: This is the beginnings of my generic create stuff.  It's not used in this PR, but could be in the future for subjects and methods.

src/helpers/cpdRoutes.js
```diff
@@ -10,6 +10,7 @@ export class CpdRoutes {
 	static get Pending() { return 'pending'; }
 	static get Progress() { return 'target/progress'; }
 	static get Question() { return 'question'; }
+	static QuestionId(questionId) { return `question/${questionId}`; }
```
TomMichielsD2L: We probably also need a MethodId, SubjectId, MethodSortOrder, QuestionSortOrder and SubjectSortOrder

src/services/cpd-service.js
```diff
@@ -12,21 +12,68 @@ d2lfetch.use({
...
+		const request = new Request(`${this.Host}${this.CpdPath(CpdRoutes.Method)}/${methodId}`, {
+			method: 'DELETE',
+			headers: {
+				'Content-Type' : 'application/json'
+			}
+		});
+		return d2lfetch.fetch(request);
+	}
+	static deleteQuestion(questionId) {
+		const request = new Request(`${this.Host}${this.CpdPath(CpdRoutes.Question)}/${questionId}`, {
```
TomMichielsD2L:
```suggestion
		const request = new Request(`${this.Host}${this.CpdPath(CpdRoutes.QuestionId(questionId)}`, {
```

src/components/cpd-admin-questions.js
```diff
@@ -1,12 +1,120 @@
...
+		this.cpdService.deleteQuestion(questionId)
+			.then(() => this.fetchQuestions());
+	}
+	fetchQuestions() {
+		return this.cpdService.getQuestions().then(data => this.questions = data);
+	}
+	openEditDialog(e) {
+		const item = e.target.getAttribute('item-json') &&
+			JSON.parse(e.target.getAttribute('item-json')) || {};
+		this.objectData = item;
```
TomMichielsD2L: why not combine?

src/components/cpd-admin-questions.js
```diff
@@ -1,12 +1,120 @@
...
+			li {
+				display: grid;
+				grid-template-columns: 30px auto;
+			}
+			`
+		];
+	}
+	constructor() {
+		super();
+		this.cpdService = CpdServiceFactory.getCpdService();
```
TomMichielsD2L: At what point should we just put this in the BaseMixin?

src/services/demo-cpd-service.js
```diff
@@ -6,12 +6,20 @@ const getOptions = {
...
+
 export class DemoCpdService {
 
 	static createRecord() {}
 
+	static createSubject() {}
+
 	static deleteRecord() {}
 
+	static deleteSubject(subjectId) {
```
TomMichielsD2L: I think we missed a few functions in this service :)

src/components/cpd-admin-questions.js
jwong-d2l: are there any places where we use localize but not the service?

src/helpers/cpdRoutes.js
jwong-d2l: I didn't want to touch method and subject yet.  I do want to deal with sortorder probably.

src/components/cpd-admin-questions.js
TomMichielsD2L: Not sure, it just feels like we have this line of code almost everywhere

BraydonKains: A couple of the small utility components. But it wouldn't be a huge detriment to just move this to the BaseMixin, I'd support that.

PR #166, opened 2020-02-18
Sortorder

jwong-d2l:


src/components/cpd-admin-questions.js
```diff
@@ -65,7 +73,7 @@ class CpdAdminQuestions extends BaseMixin(LitElement) {
 					<d2l-dropdown-content id="questionDropdown-${item.Id}">
 						<d2l-menu>
 							<d2l-menu-item item-id="${item.Id}" @click="${this.openEditDialog}" text="${this.localize('edit')}" item-json="${JSON.stringify(item)}"></d2l-menu-item>
-							<d2l-menu-item item-id="${item.Id}" @click="${this.delete}" text="${this.localize('delete')}"></d2l-menu-item>
+							<d2l-menu-item item-id="${item.Id}" ?click="${item.InUse ? null : this.delete}" text="${this.localize('delete')}" ?disabled="${item.InUse}"></d2l-menu-item>
```
BraydonKains: Does this actually work? The ? operator in lit-html is for a boolean attribute. It looks like this is still intended to be an event, just conditional whether it's null or the handler.

src/components/cpd-admin-questions.js
```diff
@@ -16,8 +16,14 @@ class CpdAdminQuestions extends BaseMixin(LitElement) {
 			questions: {
 				type: Array
 			},
+			sortable: {
```
TomMichielsD2L: Do we still need sortable, now that you are hiding the content using the showContent boolean?

src/components/cpd-admin-questions.js
jwong-d2l: It does work, but I also just got rid of it because Josh said to.

src/services/cpd-service.js
```diff
@@ -251,9 +251,15 @@ export class CpdService {
 	static get Subject() { return 'subject'; }
 	static get Target() {return 'target';}
 	static get Team() { return 'team'; }
+	static updateItemSortOrder(type, id, sortOrder) {
+		return this.putJsonRequest(`${this.CpdPath(type)}/${id}/sortorder`, {
```
TomMichielsD2L: Can we make this nicer with CpdRoutes?

src/services/demo-cpd-service.js
```diff
@@ -61,8 +61,8 @@ export class DemoCpdService {
 		const data = [
 			{Id:1, QuestionText:'Why is Ben moving?', SortOrder: 1},
 			{Id:2, QuestionText:"Why don't Tom's monitors work?"},
```
TomMichielsD2L: Do these also need single quotes?

src/components/cpd-admin-questions.js
jwong-d2l: I like disabling the sort changing anyway, for when we use the new thing, but I could take it out here.  I thought about it.

src/services/demo-cpd-service.js
BraydonKains: Yes, otherwise lint is sad

src/components/cpd-admin-questions.js
```diff
@@ -65,7 +73,7 @@ class CpdAdminQuestions extends BaseMixin(LitElement) {
 					<d2l-dropdown-content id="questionDropdown-${item.Id}">
 						<d2l-menu>
 							<d2l-menu-item item-id="${item.Id}" @click="${this.openEditDialog}" text="${this.localize('edit')}" item-json="${JSON.stringify(item)}"></d2l-menu-item>
-							<d2l-menu-item item-id="${item.Id}" @click="${this.delete}" text="${this.localize('delete')}"></d2l-menu-item>
+							${item.InUse ? html`` : html`<d2l-menu-item item-id="${item.Id}" ?click="${item.InUse ? null : this.delete}" text="${this.localize('delete')}"></d2l-menu-item>`}
```
jwong-d2l:
```suggestion
							${item.InUse ? html`` : html`<d2l-menu-item item-id="${item.Id}" 
							@click="${this.delete}" text="${this.localize('delete')}"></d2l-menu-item>`}
```

PR #170, opened 2020-02-19
De37930/hide add other user

TomMichielsD2L:


src/components/cpd-my-records.js
```diff
@@ -381,7 +381,7 @@ class MyCpdRecords extends BaseMixin(LitElement) {
 						</h2>
 						<div class="printLink">
 							<d2l-icon icon="tier1:print"></d2l-icon>
-							<d2l-link target="_blank" href="${CpdRoutes.Report(this.viewUserId)}">${this.localize('printRecords')}</d2l-link>
+							<d2l-link target="_blank" href="${CpdRoutes.UserReport(this.viewUserId)}">${this.localize('printRecords')}</d2l-link>
```
BraydonKains: Would these two `UserReport` fixes make more sense to be in the same PR?

TomMichielsD2L: Probably, but that requires changes. Might as well just merge them as is

PR #174, opened 2020-02-19
Us113704/print styles

benyamin-noori:
What it looks like without print styles: 

image

What it looks like with print styles: 

image

I showed these screenshots to Josh and he approved it! 👍


package.json
```diff
@@ -16,7 +16,6 @@
       "loader": "iframe"
     },
     "frauLocalAppResolver": {
-      "hostName": "localhost",
```
jwong-d2l: this doesn't actually matter... we can leave it in because it's always only the local appresolver.

benyamin-noori: I know, it's there. Mine was a duplicate that I accidentally committed.

src/components/progress-overall.js
```diff
@@ -90,18 +93,34 @@ class ProgressOverall extends BaseMixin(LitElement) {
...
 					align-items: center;
 					justify-items: center;
 				}
 				.progress-inner-chevron {
 					display: inline-block;
 				}
+				.progress-split-text {
+					display: none;
+				}
+				.progress-split-text {
```
jwong-d2l: this is duplicated?

src/components/progress-overall.js
```diff
@@ -171,6 +190,44 @@ class ProgressOverall extends BaseMixin(LitElement) {
 				</div>
 			</div>
 		</div>
+		<div class="progress-inner-print">
```
jwong-d2l: this will work... the only reason I'm hesitant about duplicating the progress div just for printing is that if we ever need to adjust the radials, it's going to be changes in two places, and it's not intuitive to check the print version as well.  Was there a strong reason for not just doing it inline?

src/components/progress-overall.js
benyamin-noori: Yep. Good catch.

src/components/progress-overall.js
benyamin-noori: That's a great point. I did try to use the same HTML for both, but it gets really messy and difficult to make changes to later. At some point, I realized it was getting hard for me to track the CSS in normal and print mode, so I redid it as a separate component. 

What you're saying is valid, though. I still think we _can_ do it as one but it wouldn't be great. What do you think?

BraydonKains: I agree with Joseph, I think we'd rather not do all this code duplication even if the css gets a little more complicated.

src/components/progress-overall.js
```diff
@@ -82,6 +83,12 @@ class ProgressOverall extends BaseMixin(LitElement) {
 				justify-content: center;
 				align-items: center;
 			}
+			.progress-inner-print {
```
TomMichielsD2L: we don't need no progress-inner-print. We don't need no thought control

src/components/progress-overall.js
```diff
@@ -155,6 +171,9 @@ class ProgressOverall extends BaseMixin(LitElement) {
 				<d2l-icon icon="tier3:chevron-right"></d2l-icon>
 			</div>
 			<div class="progress-inner-summary">
+				<div class="d2l-body-compact print">
```
jwong-d2l: print is probably too generic... I like the look of this, but it's probably safer to make it its own class.

benyamin-noori: Done. Renamed to `target-progress-print-header`.

PR #177, opened 2020-02-24
Changed the validation pattern on the add item dialog

BraydonKains:
empty_attempt

### DIFF
--- a/src/components/cpd-admin-items.js
+++ b/src/components/cpd-admin-items.js
@@ -27,9 +27,6 @@ class CpdAdminItems extends BaseMixin(LitElement) {
 			},
 			showContent: {
 				type: Boolean
-			},
-			validName: {
-				type: Boolean
 			}
 		};
 	}
@@ -98,8 +95,7 @@ class CpdAdminItems extends BaseMixin(LitElement) {
 	save() {
 		const objectNameInput = this.shadowRoot.querySelector('#objectName');
 		const objectName = objectNameInput.value;
-		this.validate(objectName);
-		if (this.validName) {
+		if (this.validate(objectName)) {
 			if (!this.objectData.Id) {
 				this.objectData[this.context.textFieldName] = objectName;
 				this.objectData.SortOrder = 0;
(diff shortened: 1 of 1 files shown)


src/components/cpd-admin-items.js
```diff
@@ -92,18 +96,29 @@ class CpdAdminItems extends BaseMixin(LitElement) {
...
+				this.objectData.SortOrder = 0;
+				this.cpdService.Create(this.context.type)(this.objectData)
+					.then(() => this.fetchItems()
+						.then(() => this.objectData = {}));
+			} else {
+				this.cpdService.Update(this.context.type)(this.objectData.Id)(this.objectData)
+					.then(() => this.fetchItems());
+			}
+			objectNameInput.setAttribute('aria-invalid', false);
+			this.shadowRoot.querySelector(`#create-${this.context.type}-dialog`)._close();
```
BraydonKains: We might want to call this tech debt... Apparently the dialog may be improved to handle something like this. For now we're using a secret function and it's not great. But I cannot seem to think of another way to manually close the dialog.

jwong-d2l: Maybe create it as an issue on that repo?

src/components/cpd-admin-items.js
```diff
@@ -92,18 +96,29 @@ class CpdAdminItems extends BaseMixin(LitElement) {
...
-
-		if (!this.objectData.Id) {
-			this.objectData.SortOrder = 0;
-			this.cpdService.Create(this.context.type)(this.objectData)
-				.then(() => this.fetchItems()
-					.then(() => this.objectData = {}));
+		const objectNameInput = this.shadowRoot.querySelector('#objectName');
+		const objectName = objectNameInput.value;
+		this.validate(objectName);
+		if (this.validName) {
```
TomMichielsD2L: This property no longer seems to be used, could we call the validate function directly here?

src/components/cpd-admin-items.js
```diff
@@ -92,18 +96,29 @@ class CpdAdminItems extends BaseMixin(LitElement) {
...
+			if (!this.objectData.Id) {
+				this.objectData[this.context.textFieldName] = objectName;
+				this.objectData.SortOrder = 0;
+				this.cpdService.Create(this.context.type)(this.objectData)
+					.then(() => this.fetchItems()
+						.then(() => this.objectData = {}));
```
BraydonKains: @jwong-d2l what does this do? And why don't we do it on Update?

jwong-d2l: clears out existing data, if you went from editing a subject, to adding a new one. and it might need to be done on update, or moved to where we open the dialog.

src/components/cpd-admin-items.js
```diff
@@ -49,6 +49,7 @@ class CpdAdminItems extends BaseMixin(LitElement) {
 		this.cpdService = CpdServiceFactory.getCpdService();
 		this.sortable = true;
 		this.showContent = true;
+		this.validName = true;
```
TomMichielsD2L: Looks like we missed removing this one.

PR #178, opened 2020-03-03
Using the valence host to build the navigation link instead of last D…

BraydonKains:
…2L page

According to the following Slack conversation https://d2l.slack.com/archives/C0PHG3QB0/p1583264038087000 the `valenceHost` property is built using the WebServerName. According to this BI report 
image
the WebServerName config variables are set to exactly what we need on SBG's instances. A couple other devs I've discussed with seem to concur that the `valenceHost` is the way to go for this as well.


src/components/cpd-widget.js
```diff
@@ -137,10 +136,8 @@ class CpdWidget extends BaseMixin(LitElement) {
 			.then((data) =>
 				this.subjects = data
 			);
-		this.cpdService.ParentHost(CpdRoutes.CpdHome())
-			.then(url => {
-				this.cpdLink = url;
-			});
+		this.cpdLink = this.cpdService.CpdHomeLink;
+		console.log(this.cpdLink);
```
TomMichielsD2L: whoops

src/services/cpd-service.js
```diff
@@ -198,13 +199,7 @@ export class CpdService {
...
 	static get JobTitle() { return 'target/jobtitles'; }
 	static get Method() { return 'method'; }
-	static ParentHost(route) {
-		return window.data.fraSettings.navigation.getLastD2LPage()
-			.then(page => {
-				const host = (new URL(page.url)).host;
-				return `http://${host}${route ? route : ''}`;
-			});
-	}
+
```
TomMichielsD2L: Do we need this empty line?

src/components/cpd-widget.js
BraydonKains: eslint why hath thou forsaken me

src/services/cpd-service.js
BraydonKains: I would love to selfishly say yes (I hate not having lines between methods) but for consistency's sake, no.

src/components/cpd-widget.js
jwong-d2l: we should check that.

PR #179, opened 2020-03-04
Refactor usages of ParentHost since that doesn't exist anymore

BraydonKains:
This fixes https://rally1.rallydev.com/#/15545269080d/detail/defect/374439868384?fdp=true

I forgot that other places relied on the now-deleted `ParentHost` method.


BraydonKains:
> 
> 
> Approved assuming that `${CpdRoutes.Host}${recordRoute}` results in a properly constructed path

If the concern is a `host//recordRoute` or something along those lines, I have tested and confirmed that doesn't happen.

PR #184, opened 2020-03-23
Progress on homepage respects filters

BraydonKains:
I isolated the work I did to get the progress route working with filters so I can get that in as well as the fix for the bug Gauth found (which I'm pretty sure was fixed before but probably just hasn't been released yet).

### DIFF
--- a/src/components/cpd-pending-records.js
+++ b/src/components/cpd-pending-records.js
@@ -60,9 +60,9 @@ class PendingRecords extends BaseMixin(LitElement) {
 		this.page = 1;
 		this.hideSearchOptions = true;
 		this.filters = {
-			Name: 0,
-			StartDate: 0,
-			EndDate: 0
+			Name: null,
+			StartDate: null,
+			EndDate: null
 		};
 	}
 
--- a/src/components/cpd-pending-records.js
+++ b/src/components/cpd-pending-records.js
@@ -59,6 +59,11 @@ class PendingRecords extends BaseMixin(LitElement) {
 		this.cpdService = CpdServiceFactory.getCpdService();
 		this.page = 1;
 		this.hideSearchOptions = true;
+		this.filters = {
+			Name: 0,
+			StartDate: 0,
+			EndDate: 0
+		};
 	}
 
 	connectedCallback() {
@@ -78,7 +83,6 @@ class PendingRecords extends BaseMixin(LitElement) {
 	}
 
 	updateFilter(e) {
-		this.filters = this.filters || {};
 		switch (e.target.id) {
 			case 'search_input':
 				this.filters.Name = e.detail;
--- a/src/components/cpd-record-report.js
+++ b/src/components/cpd-record-report.js
@@ -40,6 +40,9 @@ class CpdRecordReport extends BaseMixin(LitElement) {
 			},
 			userId: {
 				type: Number
+			},
+			filters: {
+				type: Object
 			}
 		};
 	}
@@ -134,6 +137,7 @@ class CpdRecordReport extends BaseMixin(LitElement) {
 		this.records = [];
 		this.questions = [];
 		this.progress = {};
+		this.filters = {};
 		this.types = this.cpdService.getTypes();
 	}
 
--- a/src/services/cpd-service.js
+++ b/src/services/cpd-service.js
@@ -78,54 +78,54 @@ export class CpdService {
 		return this.getRequest(CpdRoutes.RelativePath(type));
 	}
 	static getJobTitleDefaults(page) {
-		let api_path = CpdRoutes.RelativePath(this.JobTitle);
-		api_path += `?pageNumber=${page}`;
-		return this.getRequest(api_path);
+		let url = CpdRoutes.RelativePath(this.JobTitle);
+		const searchParams = new URLSearchParams();
+		searchParams.append('pageNumber', page);
+		url = url.concat(`?${searchParams.toString()}`);
+		return this.getRequest(url);
 	}
 	static getMethods() {
 		return this.getRequest(CpdRoutes.RelativePath(this.Method));
 	}
 	static getMyTeam(page, filters) {
-		let api_path = CpdRoutes.RelativePath(this.Team);
-		api_path += `?pageNumber=${page}`;
+		let url = CpdRoutes.RelativePath(this.Team);
+		const searchParams = new URLSearchParams();
+		searchParams.append('pageNumber', page);
 		if (filters) {
 			const { Name } = filters;
-			if (Name && Name.value) api_path += `&searchTerm=${encodeURIComponent(Name.value)}`;
+			if (Name && Name.value) searchParams.append('searchTerm', Name.value);
 		}
-		return this.getRequest(api_path);
+		url = url.concat(`?${searchParams.toString()}`);
+		return this.getRequest(url);
 	}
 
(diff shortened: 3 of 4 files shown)


src/services/cpd-service.js
```diff
@@ -107,11 +107,25 @@ export class CpdService {
 		return this.getRequest(api_path);
 	}
 
-	static getProgress(userId) {
-		if (!userId) {
-			return this.getRequest(CpdRoutes.RelativePath(CpdRoutes.Progress));
+	static getProgress(filters, userId) {
```
jwong-d2l: as written, this will break current behaviour... I think you want 
```suggestion
	static getProgress(userId, filters) {
```

src/services/cpd-service.js
```diff
@@ -170,11 +184,25 @@ export class CpdService {
 		}
 		return this.getRequest(CpdRoutes.RelativePath(CpdRoutes.JobTarget(jobTitle)));
 	}
-	static getTargetRecords(userId) {
-		if (!userId) {
-			return this.getRequest(CpdRoutes.RelativePath(CpdRoutes.ReportRecords));
+	static getTargetRecords(filters, userId) {
+		let api_path = userId ? CpdRoutes.RelativePath(CpdRoutes.UserReportRecords(userId)) : CpdRoutes.RelativePath(CpdRoutes.ReportRecords);
```
jwong-d2l: You really like these underscores?  CamelCase for consistency please!

src/services/cpd-service.js
```diff
@@ -170,11 +184,25 @@ export class CpdService {
...
+		if (filters) {
+			api_path += '?';
+			const query_params = [];
+			const { Subject, Method, Name, StartDate, EndDate } = filters;
```
jwong-d2l: would something like the below be readable?
Object.keys(filters)
.filter( (filter) => filters[filter].enabled === true)
.map( (filter) => `${filter}=${filters[filter].value}`)
.join('&');

src/services/cpd-service.js
```diff
@@ -107,11 +107,25 @@ export class CpdService {
...
+			if (Method.value && Method.enabled) query_params.push(`method=${Method.value}`);
+			if (Name.value) query_params.push(`recordName=${encodeURIComponent(Name.value)}`);
+			if (StartDate.value) query_params.push(`startDate=${dateParamString(StartDate.value)}`);
+			if (EndDate.value) query_params.push(`endDate=${dateParamString(EndDate.value, true)}`);
+			for (let i = 0; i < query_params.length; i++) {
+				if (i > 0) {
+					api_path += '&';
+				}
+				api_path += query_params[i];
+			}
```
TomMichielsD2L: Could we either join the queryParams array with an ampersand, or use something like the URL object to just add the params as we find them. I think I like the second option more.
`api_path += query_params.join('&');`
OR
`
if (filters) {
const url = new URL(api_path);
if (Subject.value && Subject.enabled) url.searchParams.append('subject',Subject.value);
....
api_path = url.href;
}
`

TomMichielsD2L: Looks like Joseph has a similar comment down below :D

src/services/cpd-service.js
TomMichielsD2L: Perhaps we can combine this with the whole url.searchParams.append() logic.

src/components/cpd-pending-records.js
```diff
@@ -59,6 +59,11 @@ class PendingRecords extends BaseMixin(LitElement) {
 		this.cpdService = CpdServiceFactory.getCpdService();
 		this.page = 1;
 		this.hideSearchOptions = true;
+		this.filters = {
+			Name: 0,
+			StartDate: 0,
+			EndDate: 0
+		};
```
TomMichielsD2L: Does this result in us filtering name by 0 and such if there are no actual filters selected?

src/components/cpd-record-report.js
```diff
@@ -40,6 +40,9 @@ class CpdRecordReport extends BaseMixin(LitElement) {
 			},
 			userId: {
 				type: Number
+			},
+			filters: {
+				type: Object
```
TomMichielsD2L: To be populated in the future?

src/services/cpd-service.js
```diff
@@ -78,40 +78,54 @@ export class CpdService {
 		return this.getRequest(CpdRoutes.RelativePath(type));
 	}
 	static getJobTitleDefaults(page) {
-		let api_path = CpdRoutes.RelativePath(this.JobTitle);
-		api_path += `?pageNumber=${page}`;
-		return this.getRequest(api_path);
+		let url = CpdRoutes.RelativePath(this.JobTitle);
```
TomMichielsD2L: Should we make this return an actual url object, since we seem to be moving towards using the url related objects?

src/components/cpd-record-report.js
BraydonKains: I didn't mean to push this. I guess it's technically harmless for now since it's just unused.

src/components/cpd-pending-records.js
BraydonKains: No, we check that Name has a truthy property called `value` (i.e. it's been set through the UI and isn't empty)

src/services/cpd-service.js
BraydonKains: As it is right now we can't create a URL object because it's only the relative path. It does deserve a refactor, but I had to stop the refactor somewhere for this PR. All of our filter stuff is not as clean as it could be.

src/components/cpd-pending-records.js
TomMichielsD2L: gotcha, I would prefer we assign them to null as the default.

src/services/cpd-service.js
TomMichielsD2L: I like the idea of dynamically going through the filters, it would allow this to scale as needed. Perhaps a foreach instead of a filter, fiven the new searchParams stuff.

BraydonKains: This isn't going to be as easy as it looks, since right now our different routes have slightly differently named query parameters. We would need consistent naming on the back end before we can make that refactor.

jwong-d2l: The filters are different types of values too, in that some of them need to be transformed or pushed to end.

src/services/cpd-service.js
```diff
@@ -78,40 +78,54 @@ export class CpdService {
...
-		api_path += `?pageNumber=${page}`;
+		let url = CpdRoutes.RelativePath(this.Pending);
+		const searchParams = new URLSearchParams();
+		searchParams.append('pageNumber', page);
 		if (filters) {
 			const { Name, StartDate, EndDate } = filters;
-			if (Name && Name.value) api_path += `&awardName=${encodeURIComponent(Name.value)}`;
-			if (StartDate && StartDate.value) api_path += `&startDate=${dateParamString(StartDate.value)}`;
-			if (EndDate && EndDate.value) api_path += `&endDate=${dateParamString(EndDate.value, true)}`;
+			if (Name.value) searchParams.append('recordName', Name.value);
```
jwong-d2l: If we leave it as Name && Name.value, does that mean we can remove the dummy filter object instantiation?

BraydonKains: I initially switched it to this for consistency with everything else, but the old way was probably better.

PR #188, opened 2020-03-26
Us113138/csv export

benyamin-noori:
Adds a dropdown menu instead of the print button that has two items: 

1. Print records. 
2. Export CSV.


package.json
```diff
@@ -77,7 +77,6 @@
     "@webcomponents/webcomponentsjs": "^2",
     "d2l-date-picker": "github:BrightspaceUI/date-picker#semver:^2",
     "d2l-dnd-sortable": "Brightspace/dnd-sortable#semver:^3",
-    "d2l-dropdown": "BrightspaceUI/dropdown#semver:^7",
```
jwong-d2l: thanks for doing this

src/components/cpd-my-records.js
```diff
@@ -369,8 +387,14 @@ class MyCpdRecords extends BaseMixin(LitElement) {
 							${this.localize('userTitle', { 'userName': this.userDisplayName})}
 						</h2>
 						<div class="printLink">
-							<d2l-icon icon="tier1:print"></d2l-icon>
-							<d2l-link target="_blank" href="${this.printRecordLink}">${this.localize('printRecords')}</d2l-link>
+							<d2l-dropdown-button primary text="${this.localize('printExportRecords')}">
+								<d2l-dropdown-menu>
+									<d2l-menu>
+										<d2l-menu-item-link text="${this.localize('printRecords')}" id="print_records"></d2l-menu-item-link>
```
jwong-d2l: did it not work with the href property?  It also has a secret (read: undocumented) target property.

jwong-d2l: https://github.com/BrightspaceUI/core/blob/eb103a9e03e5c8c495273eaf5a4bba56d92c98e6/components/menu/menu-item-link.js#L10

jwong-d2l: I know that for the csv you probably can't finagle that to work, but it bothers me like crazy when I can't right-click a link and then open in new tab.

src/services/cpd-service.js
```diff
@@ -74,6 +74,28 @@ export class CpdService {
...
 
+	static getCsvExport(userId) {
+		let path;
+		if (userId) {
+			path = `${this.Host}${CpdRoutes.RelativePath(CpdRoutes.UserCSVReport(userId))}`;
+		} else {
+			path = `${this.Host}${CpdRoutes.RelativePath(CpdRoutes.CSVReport)}`;
+		}
+		const request = new Request(path);
+		return d2lfetch.fetch(request)
```
jwong-d2l: We might be able to avoid this if we can get @TomMichielsD2L to do something like https://search.d2l.dev/xref/lms/le/assessments/D2L.LE.Assessments/Integration/Web/CSVGradesController.cs?r=13c5110f#45 on the controller endpoint

jwong-d2l: Actually, tom already returns an IFileActionResult, so we should be able to direct link to the file without the fetching.

src/components/cpd-my-records.js
benyamin-noori: Hmm... good point. The reason I didn't use `href` was because clicking on it wouldn't open a new tab, it would change the current window's URL to the `href` and that meant the iFrame's URL would change and the LMS header stuff would still be there. 

I'll see if I can do both, have an `href` for right clicking and "opening in new tab" and also even handling for left clicks.

TomMichielsD2L: The secret target property should allow you to open it in a new tab.

src/components/cpd-my-records.js
```diff
@@ -148,13 +152,30 @@ class MyCpdRecords extends BaseMixin(LitElement) {
...
-				.then(data => {
-					this.userDisplayName = data;
-				});
-		}
+		this.cpdService.getUserInfo(this.viewUserId)
+			.then(data => {
+				this.userDisplayName = data;
+			});
 		this.updatePrintRecordLink();
+		this.addEventListener('d2l-menu-item-select', this.d2lMenuItemSelectListener);
```
TomMichielsD2L: I think for all other events we define the listener as an attribute in the html tag. Did that not work here?

src/components/cpd-my-records.js
jwong-d2l: blegh, file downloads are an api call that need a bearer token.

PR #190, opened 2020-03-30
Us113817/report uses filters

BraydonKains:
The report will now respect the filters coming from the My Cpd Records page.


src/components/cpd-record-report.js
```diff
@@ -241,7 +265,7 @@ class CpdRecordReport extends BaseMixin(LitElement) {
 						</div>
 						<div>
 							<span class="d2l-heading-3">${this.localize('cpdPeriod')}</span>
-							${this.localize('range', this.target)}
+							${this.target ? this.localize('range', this.target) : ''}
```
TomMichielsD2L: FYI: We know that this will not work if only the startDate or endDate is specified. We are waiting for feedback from Josh as to what the langterm should say in that case.

src/components/cpd-record-report.js
```diff
@@ -137,22 +149,34 @@ class CpdRecordReport extends BaseMixin(LitElement) {
 		this.records = [];
 		this.questions = [];
 		this.progress = {};
-		this.filters = {};
+		this.subject = 0;
+		this.method = 0;
+		this.recordName = '';
+		this.endDate = '';
+		this.startDate = '';
```
TomMichielsD2L: Does this initialization do anything for us?

src/components/cpd-record-report.js
```diff
@@ -223,7 +247,7 @@ class CpdRecordReport extends BaseMixin(LitElement) {
 				<div>
 					${unsafeHTML(answer.Text)}
 				</div>
-			<div>
+			</div>
```
jwong-d2l: wow

src/components/cpd-record-report.js
BraydonKains: We've had inconsistent conclusions when it comes to whether this actually does anything for us, but I've always done it because the website says to. https://lit-element.polymer-project.org/guide/properties#initialize

jwong-d2l: Is it an implicit "initialize them here", if you need to initialize them?

jwong-d2l: The example never sets them to anything empty or null

src/components/cpd-record-report.js
```diff
@@ -241,7 +280,7 @@ class CpdRecordReport extends BaseMixin(LitElement) {
 						</div>
 						<div>
 							<span class="d2l-heading-3">${this.localize('cpdPeriod')}</span>
-							${this.localize('range', this.target)}
+							${this.target ? this.renderRangeString() : ''}
```
TomMichielsD2L: Can we combine the function and this null check?

src/components/cpd-record-report.js
TomMichielsD2L: So, we should not be doing this?

BraydonKains: It's possible I've been misunderstanding it all this time. I'll try taking them out.

src/components/cpd-record-report.js
BraydonKains: oops I forgot to take that out here

PR #195, opened 2020-04-06
Us112465/dismiss pending

TomMichielsD2L:
image

Updated pending records, to be able to dismiss a pending record. Current implementation makes assumptions about the api.


locales/en.json
```diff
@@ -135,6 +137,7 @@
   "to": "to",
   "total": "Total",
   "type": "Type",
+  "undismissRecord" : "Undismiss Record",
```
gravindran-d2l, line 140: Maybe clear with Josh, but "undismiss" does not seem like a good word.

TomMichielsD2L: I agree :D

src/components/cpd-pending-records.js
```diff
@@ -110,12 +114,25 @@ class PendingRecords extends BaseMixin(LitElement) {
...
 	}
 
+	dismissRecordButtonClicked(e) {
+		const awardData = JSON.parse(e.target.getAttribute('data-award-id'));
+		if (this.filters.view === 'dismissed') {
+			this.cpdService.undismissRecord(awardData.IssuedAwardId);
+		}
+		else {
+			this.cpdService.dismissRecord(awardData.IssuedAwardId);
+		}
```
TomMichielsD2L, line 133: Maybe we should re-fetch the pending at this point?

locales/en.json
TomMichielsD2L: Restore is the new word. It is a better word.

src/helpers/cpdRoutes.js
```diff
@@ -1,6 +1,7 @@
 export class CpdRoutes {
 	static CpdHome() { return '/d2l/custom/cpd/main'; }
 	static get CSVReport() { return 'record/report/csv'; }
+	static DismissRecord(awardId) { return `pending/dismiss${awardId}`; }
```
jwong-d2l, line 4:
```suggestion
	static DismissRecord(awardId) { return `pending/dismiss/${awardId}`; }
```

PR #196, opened 2020-04-07
US115052 making sure the click function completes before we revoke th…

TomMichielsD2L:
…e URL


src/services/demo-cpd-service.js
```diff
@@ -28,6 +28,29 @@ export class DemoCpdService {
...
+		document.body.appendChild(a);
+		a.click().then(() => {
+			window.URL.revokeObjectURL(url);
+			document.body.removeChild(a);
+		});
+	}
+
+	static async getAttachment(attachment) {
+		let blob;
+		if (attachment.href) {
```
TomMichielsD2L, line 46: I can only test in edge via the demo, and do not need the service to try to download from the server

PR #198, opened 2020-04-08
CSV Export on Edge

benyamin-noori:


src/services/demo-cpd-service.js
```diff
@@ -41,14 +46,15 @@ export class DemoCpdService {
...
-		}
-		this.downloadBlob(blob, attachment.name);
+	static async getAttachment(fileName) {
+		const blob = new Blob([JSON.stringify({key: 'value'})], {type : 'application/json'});
+		this.downloadBlob(blob, fileName);
+	}
+
+	static getCsvExport() {
+		const now = new Date();
+		const fileName = `${this.userDisplayName}_${now.toLocaleDateString()} ${now.toLocaleTimeString()}.csv`;
```
gravindran-d2l, line 56: Will that put a space in the file name between the time and the date?

benyamin-noori: Yes, why?

gravindran-d2l: Never mind, I thought there was originally an underscore between the time and date, but I see that it was not the case.  Carry on!

PR #200, opened 2020-04-13
US115052 Making sure that edge does not download multiple files, and …

TomMichielsD2L:
…that chrome does not throw


src/services/cpd-service.js
```diff
@@ -88,6 +88,7 @@ export class CpdService {
 		let url;
 		if (window.navigator.msSaveOrOpenBlob) {
 			url = window.navigator.msSaveOrOpenBlob(blob, fileName);
```
TomMichielsD2L, line 90: According to https://docs.microsoft.com/en-us/previous-versions/hh772332(v%3Dvs.85)
Save or open, makes the browser open the file, it also works that way in my tests.

src/services/cpd-service.js
```diff
@@ -96,10 +97,9 @@ export class CpdService {
 		a.href = url;
 		a.download = fileName;
 		document.body.appendChild(a);
-		a.click().then(() => {
-			window.URL.revokeObjectURL(url);
-			document.body.removeChild(a);
-		});
+		a.click();
```
TomMichielsD2L, line 100: This is no longer reached by edge, and not async in the chrome world

src/services/demo-cpd-service.js
```diff
@@ -42,21 +43,26 @@ export class DemoCpdService {
...
+		this.downloadBlob(blob, attachment.name);
 	}
 
 	static getCsvExport() {
 		const now = new Date();
-		const fileName = `${this.userDisplayName}_${now.toLocaleDateString()} ${now.toLocaleTimeString()}.csv`;
-		this.getAttachment(fileName);
+		const fileName = `${this.userDisplayName}_${now.toLocaleDateString()}_${now.toLocaleTimeString()}.csv`;
+		const blob = new Blob([JSON.stringify({key: 'value'})], {type : 'application/json'});
+		this.downloadBlob(blob, fileName);
```
TomMichielsD2L, line 65: Fixed this to be in line with the real service.

src/services/demo-cpd-service.js
```diff
@@ -42,21 +43,26 @@ export class DemoCpdService {
...
-		});
+		a.click();
+		window.URL.revokeObjectURL(url);
+		document.body.removeChild(a);
 	}
 
-	static async getAttachment(fileName) {
-		const blob = new Blob([JSON.stringify({key: 'value'})], {type : 'application/json'});
-		this.downloadBlob(blob, fileName);
+	static async getAttachment(attachment) {
```
TomMichielsD2L, line 51: This function gets called for the attachment downloads. If the attachment is saved (impossible) make on up. Otherwise just reserve the file they just added.

PR #202, opened 2020-04-20
US113446 Updated the my team page to allow the user to view the team

TomMichielsD2L:
image


src/components/cpd-my-team.js
```diff
@@ -60,28 +71,45 @@ class MyTeamCPD extends BaseMixin(LitElement) {
...
 
-		this.cpdService.getMyTeam(this.page)
-			.then(data => {
-				this.myTeam = data;
-			});
+		this.filters.ManagingUserId = this.viewUserId;
+		this.fetchData();
+	}
+
+	attributeChangedCallback(name, oldval, newval) {
```
TomMichielsD2L, line 78: I am doing this because when we navigate from someone's team, to another team. We do not get a full re-render, but the property does update.

locales/en.json
```diff
@@ -140,7 +140,9 @@
   "type": "Type",
   "unstructured": "Unstructured",
   "userDetails": "User Details",
+  "userTeam": "{userName}'s Team CPD",
```
jwong-d2l, line 143: should we update this to userDisplayName?  It's not really a username.

TomMichielsD2L: I feel like userDisplayName is more of a back end concept. Our front end does not really differentiate. But I can change this if you feel strongly?

jwong-d2l: I don't, you can leave it.

TomMichielsD2L: Not sure why I pushed back on this.. But we might as well get the terms in sync, so I updated it.

PR #210, opened 2020-05-28
Updating Translations 2006 28/05/2020

BrightspaceGitHubReader:


josephsaffron:
IS it just me or is this just setting some things to english?

TomMichielsD2L:
> IS it just me or is this just setting some things to english?

Sure looks like that. The only explanation I have is that we maybe changed these langterms recently and the current translations were out of date.

PR #211, opened 2020-06-29
Fix startdate and enddate filter processing

BraydonKains:
Well that's embarrassing. I forgot that the format for StartDate and EndDate change event is different for the new date picker and didn't update the service to accommodate that. So it was always getting essentially blank dates. 
For the new date picker we just pass the chosen value instead of the detail of the event (since it's no longer using a custom event, rather the browser-native change event).

### DIFF
--- a/src/services/cpd-service.js
+++ b/src/services/cpd-service.js
@@ -159,8 +159,8 @@ export class CpdService {
 		if (filters) {
 			const { Name, StartDate, EndDate, Dismissed } = filters;
 			if (Name && Name.value) searchParams.append('awardName', Name.value);
-			if (StartDate && StartDate.value) searchParams.append('startDate', dateParamString(StartDate.value));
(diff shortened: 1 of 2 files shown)


src/services/cpd-service.js
```diff
@@ -307,8 +307,8 @@ export class CpdService {
 			if (Subject && Subject.value && Subject.enabled) searchParams.append('subject', Subject.value);
 			if (Method && Method.value && Method.enabled) searchParams.append('method', Method.value);
 			if (Name && Name.value) searchParams.append('recordName', Name.value);
-			if (StartDate && StartDate.value) searchParams.append('startDate', dateParamString(StartDate.value));
-			if (EndDate && EndDate.value) searchParams.append('endDate', dateParamString(EndDate.value, true));
+			if (StartDate) searchParams.append('startDate', dateParamString(StartDate));
```
BraydonKains, line 310: Not sure why github is showing newlines here, they aren't in the actual file.

PR #214, opened 2020-07-14
Fixing Adjust Targets Conditional

BraydonKains:
I don't know how this took 4 tries to straighten out in my head, but I think it's finally the correct condition.

### DIFF
--- a/src/components/progress-box.js
+++ b/src/components/progress-box.js
@@ -79,7 +79,7 @@ class CpdProgressBox extends BaseMixin(LitElement) {
 	}
 
 	renderTargetLink() {
-		if (!this.viewUserId || !this.hasEnforcedTarget) {
+		if (!this.viewUserId && !this.hasEnforcedTarget) {
 			return html`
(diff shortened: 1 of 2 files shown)


src/components/progress-box.js
```diff
@@ -79,7 +79,7 @@ class CpdProgressBox extends BaseMixin(LitElement) {
 	}
 
 	renderTargetLink() {
-		if (!this.viewUserId || !this.hasEnforcedTarget) {
+		if (!this.viewUserId && !this.hasEnforcedTarget) {
```
TomMichielsD2L, line 82: Truth table time:
viewUserId | hasEnforcedTarget | result
      T                     T                          F
      T                     F                          F
      F                     T                          F
      F                     F                          T
Looks like we have it figured out this time.

BraydonKains: I should've done that from the start, I ended up using the strategy of reading the condition in english to see if it was right 😄

PR #215, opened 2020-07-14
LitElement doesn't like conditionally rendering attributes

BraydonKains:
Changed strategies so that we're not passing an undefined attribute but we're not conditionally rendering them.

### DIFF
--- a/src/components/cpd-my-records.js
+++ b/src/components/cpd-my-records.js
@@ -138,8 +138,6 @@ class MyCpdRecords extends BaseMixin(LitElement) {
 		this.filters = {};
 
 		this.hideSearchOptions = true;
-
-		this.viewUserId = 0;
 	}
 
 	connectedCallback() {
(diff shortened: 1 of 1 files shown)


src/components/cpd-my-records.js
```diff
@@ -408,8 +410,8 @@ class MyCpdRecords extends BaseMixin(LitElement) {
 				${this.renderHeader(this.viewUserId)}
 				<d2l-cpd-progress-box
 					?hasEnforcedTarget=${this.hasEnforcedTarget}
-					${ this.viewUserId ? html`viewUserId=${this.viewUserId}` : null }
 					.progress="${this.progress}"
+					viewUserId=${this.viewUserId}
```
gravindran-d2l, line 414: Can you do "this.viewUserId || 0" here?

TomMichielsD2L: alternatively, is lit-element happier if instead of returning null you return `html```?

BraydonKains: Yeah that's better, I'll do that instead.

PR #222, opened 2021-01-19
GitHub actions

TomMichielsD2L:
Github actions instead of Travis


BraydonKains (approved):

julianpettit (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
+name: CI
+on: pull_request
+
+jobs:
+  test:
+    name: Test
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Brightspace/third-party-actions@actions/checkout
```
josephsaffron, line 10: should you be slotting the uses under the step that uses it?  Or does it not matter?

josephsaffron: Or I guess this is just an unnamed step

TomMichielsD2L: I will name these steps

PR #231, opened 2021-03-17
OSLO integration - update serge file and translation mixin

imcdonald-d2l:
Updating serge file collection name to remove spaces. See [slack thread here](https://d2l.slack.com/archives/C01ATGP7A5A/p1615411514003200) on why this decision was made. 

Updating translation mixin so it aligns with OSLO specs [defined here.](https://docs.dev.d2l/index.php/Oslo#Create_a_mixin_to_expose_the_localize.28.29_method) 

Updating `package-lock.json` so we get latest version of dependencies. `@brightspace-ui/core` was out of date.


src/mixins/base-mixin.js
```diff
@@ -1,50 +1,57 @@
...
 		}
 
-		return null;
+		const response = await fetch(getLangUrl('en'));
+		const translations = await response.json();
+		return await getLocalizeOverrideResources(
+			'en',
+			translations,
+			resolveOverridesFunc
+		);
```
imcdonald-d2l, line 54: This is a fail safe in case all `let response = await fetch(langUrl);` return `200` but `translations = await response.json();` doesn't return anything. Unlikely to actually happen but looks like it is standard practice.

lweber-d2l: Yay for defensive programming! 👍

lweber-d2l (approved):

PR #236, opened 2021-03-23
Fix font issue so it uses Lato - us125032

imcdonald-d2l:
Followed instructions [here](https://github.com/BrightspaceUI/core/tree/master/components/typography#app) to import D2L font styles within our FRA so they are rendered properly. 

Before and after screenshots are provided below. Changes are most notable when looking at the `g` character. 

**Widget Before**
font-broken

**Widget After**
font-fix

**CPD Before**
font-broken-2

**CPD After**
font-fix-2

Rally ticket: https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F499641959940


polymer.json
```diff
@@ -16,7 +16,8 @@
   },
   "extraDependencies": [
     "node_modules/@webcomponents/**",
-    "node_modules/d2l-html-editor/**"
+    "node_modules/d2l-html-editor/**",
+    "node_modules/@brightspace-ui/core/components/typography/**"
```
imcdonald-d2l, line 20: Had to add this to `extraDependencies` since `polymer build` command only keeps dependencies added to any files from `src/...` by default. Since new dependency was added to `index.html` at the root level, we have to explicitly add dependencies here.

image

Syeda-Azim (approved):

PR #242, opened 2021-03-26
Localize page title - us126282

imcdonald-d2l:
Update page title with localized version if it differs from the default "Continuous Professional Development". Add handling for when translation for term does not yet exist.

This will have OSLO support with new lang term. We won't be able to test with translations yet since we need to wait for serge to get those in (looks like serge gets kicked off every two weeks - you can check #serge_status slack channel). 

https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F515086267872


src/mixins/base-mixin.js
```diff
@@ -80,7 +80,7 @@ export const BaseMixin = superclass => class extends RtlMixin(LocalizeMixin(supe
 	}
 
 	localize(key, params) {
-		return super.localize(key, params) || `{language term '${key}' not found}`;
+		return super.localize(key, params) || "";
```
imcdonald-d2l, line 83: Don't know why this was the decided approach when lang terms don't exist... pretty bad UX for the end-user. User shouldn't see glaring errors in the UI:
image

This would show when switching languages since serge has not yet given us translations yet. Changing it to `""` so that we can check if translations exist before updating the tab title.

src/components/cpd-admin.js
```diff
@@ -33,6 +33,11 @@ class CpdAdmin extends BaseMixin(LitElement) {
 		this.pageData = e.detail.pageData;
 	}
 	render() {
+		const pageTitle = this.localize('cpd:title');
+		if (pageTitle && document.title !== pageTitle) {
+			document.title = pageTitle;
+		}
```
imcdonald-d2l, line 39: In javascript `""` evaluates to false. When translations don't exist we will keep the default `Continuous Professional Development` tab title.

src/mixins/base-mixin.js
```diff
@@ -80,7 +80,7 @@ export const BaseMixin = superclass => class extends RtlMixin(LocalizeMixin(supe
 	}
 
 	localize(key, params) {
-		return super.localize(key, params) || `{language term '${key}' not found}`;
+		return super.localize(key, params) || '';
```
imcdonald-d2l, line 83: Don't know why this was the decided approach when lang terms don't exist... pretty bad UX for the end-user. User shouldn't see glaring errors in the UI:
image

This would show when switching languages since serge has not yet given us translations yet. Changing it to `""` so that we can check if translations exist before updating the tab title.

lweber-d2l (approved):

PR #244, opened 2021-03-29
Fix structured/unstructured text missing translations - DE43012

imcdonald-d2l:
image

https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F515484181088


src/mixins/base-mixin.js
```diff
@@ -1,60 +1,7 @@
...
-					translations,
-					resolveOverridesFunc
-				);
-			}
-		}
-
-		return await defaultTranslations();
-	}
+import { LocalizationMixin } from './localization-mixin';
+export const BaseMixin = superclass => class extends RtlMixin(LocalizationMixin(superclass)) {
```
imcdonald-d2l, line 4: Refactored mixin to remove all localization pieces into it's own mixin (`LocalizationMixin`) for cleaner code separation. Extended the base mixin to include `LocalizationMixin` to get the logic that was moved. No logic has been changed, it has just been moved to a new file.

src/mixins/localization-mixin.js
```diff
@@ -0,0 +1,71 @@
...
+	getLocalizedTypes() {
+		return [ {
+			Id: 1,
+			Name: this.localize('shared:structuredText')
+		},
+		{
+			Id: 0,
+			Name: this.localize('shared:unstructuredText')
+		}];
+	}
```
imcdonald-d2l, line 70: Replaced `getTypes()` in `cpd-service` with `getLocalizedTypes()` here so we can get translated types. Can't extend `cpd-service` with `LocalizationMixin` to get translations since it is not a `LitElement` component, it is just a helper class.

src/components/cpd-add-record.js
```diff
@@ -255,6 +255,7 @@ class AddCpdRecord extends BaseMixin(LitElement) {
 	}
 
 	render() {
+		this.types = this.getLocalizedTypes();
```
imcdonald-d2l, line 258: `this.getLocalizedTypes()` can't be called in the constructor since `this.localize()` can only be called in the `render()` function.

lweber-d2l (approved):
Nice solution! 👍

PR #246, opened 2021-03-30
CIPD Preliminary UI Changes

bamcdole:
This branch will serve as a "starting point" that contains the UI changes that are common to all of the individual UI cases in Rally.

We are using this as a temporary "master" branch for the various UI cases, so that we can parallelize work before the LMS-side data model changes are finished.


src/components/cpd-admin.js
```diff
@@ -11,6 +11,13 @@ class CpdAdmin extends BaseMixin(LitElement) {
 		return {
 			context: {
 				type: Object
```
imcdonald-d2l, line 13: This would probably be better off as three separate properties instead so we know exactly what the component needs instead of an ambiguous object. Would also make setting properties nicer so we don't have to do this: `context="{&quot;admin&quot;:true,&quot;metadata&quot;:true,&quot;trackingMethod&quot;:1}"`

bamcdole: I figured out where the difference in behavior occurs: https://search.d2l.dev/xref/Brightspace/continuous-professional-development/index.html?r=8e464a78#25-51

Looks like the way Free Range Apps work is that a root HTML document is hosted within Brightspace, and there is some logic that occurs up front which processes the incoming FRA-related data, one piece of which is the Dictionary of attributes we define within the LMS.

The dictionary is what becomes our 'frauSettings', and this page uses those settings to create the desired component _differently_, depending on the needs of the component.

imcdonald-d2l: Ah interesting, thanks for adding your findings! I'm not set on making sure we modify how the admin page gets the necessary context information but it would be nice if there could be consistency in how the main component and admin page are loaded.

lweber-d2l (approved):
Looks good so far 👍

demo/index.html
```diff
@@ -26,12 +26,13 @@ <h2>d2l-cpd</h2>
 				<d2l-cpd
 					manager
 					managePersonal
-					>
+					trackingMethod="2">
```
IhabAlkabary, line 29: Can we use the enum in `/helpers/trackingMethod`

demo/index.html
```diff
@@ -26,12 +26,13 @@ <h2>d2l-cpd</h2>
...
 				</d2l-cpd>
 			</d2l-demo-snippet>
 			<d2l-demo-snippet>
-				<d2l-cpd page="cpd-add-record"
-					>
+				<d2l-cpd
+					page="cpd-add-record"
+					trackingMethod="1">
```
IhabAlkabary, line 35: same here, can we use the enum in `/helpers/trackingMethod`

demo/index.html
bamcdole: Since these are demo page templates (as opposed to Lit Element files), we don't have access to the expression bindings that would normally allow us to reference an imported object (such as the enum).

We would have to resort to some ugly javascript in order to achieve this relatively small win, which I don't believe is worthwhile.

locales/en.json
```diff
@@ -104,7 +107,7 @@
   "shared:range": "{rangeStart} - {rangeEnd}",
   "recordReport:rangeOnlyEnd": "Records up to and including {rangeEnd}",
   "recordReport:rangeOnlyStart": "Records on or after {rangeStart}",
-  "pendingRecords:restoreRecord" : "Restore Record",
+  "pendingRecords:restoreRecord": "Restore Record",
```
Syeda-Azim, line 110: Are we removing all whitespaces from local files?

src/components/cpd-record-report.js
```diff
@@ -372,7 +374,8 @@ class CpdRecordReport extends BaseMixin(LitElement) {
 				</d2l-progress-overall>
```
Syeda-Azim, line 374: Missing `trackingMethod ` for <d2l-progress-overall>

demo/progress.html
```diff
@@ -21,8 +21,7 @@
 			<h2>d2l-cpd-progress-box</h2>
```
Syeda-Azim, line 21: Missing `trackingMethod `attribute for _d2l-progress-overall_

src/components/cpd-record-report.js
bamcdole: These will default to DurationHours if a trackingMethod isn't specified

demo/progress.html
bamcdole: These will default to DurationHours if a trackingMethod isn't specified

PR #248, opened 2021-03-30
US125040 - Update Adjust Targets page to handle 'Record Count' tracking method

bamcdole:


locales/ar-SA.json
```diff
@@ -132,5 +133,6 @@
    "myRecords:userTitle" : "سجل التطوير المهني المستمر الخاص بـ {userDisplayName}",
    "readOnlyRecord:viewCPDRecord" : "عرض سجل التطوير المهني المستمر",
    "myTeam:viewTeamCPD" : "عرض التطوير المهني المستمر للفريق",
-   "myRecords:yes" : "نعم"
+   "myRecords:yes" : "نعم",
+   "shared:records": "TODO"
```
Syeda-Azim, line 137: For new langs we just need to add it to _en.json_ and serge will auto-populate the files with translations.

PR #250, opened 2021-04-06
US126568 Pass Tracking Method to Page Components

jorge-ramirez-arredondo:
https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F517280839472


index.html
```diff
@@ -23,57 +37,65 @@
...
-			cpdWidget.setAttribute('userId', userId);
+			setAttributes(cpdWidget, { userId, trackingMethod });
 
 			const loadingSpinner = document.getElementById('loading-spinner');
 			document.body.removeChild(loadingSpinner);
 			document.body.appendChild(cpdWidget);
 		}
 		const loadCpdRecordReport = function(frauSettings) {
-			const {userId} = frauSettings;
+			const { userI, trackingMethod } = frauSettings;
```
bamcdole, line 80: Small typo here? Should this be 'userId'?

src/helpers/trackingMethodEnum.js
```diff
@@ -0,0 +1,12 @@
+const trackingMethodEnum = Object.freeze({
```
bamcdole, line 1: I had never heart of Object.freeze() before - neat!

index.html
```diff
@@ -23,57 +37,65 @@
...
 		});
 		const loadCpd = function(frauSettings) {
-			const { page, manager, managePersonal, hasEnforcedTarget } = frauSettings;
+			const {
+				page,
+				manager,
+				managePersonal,
+				hasEnforcedTarget,
+				trackingMethod
+			} = frauSettings;
```
bamcdole, line 46: Do we gain more by pulling out the individual **frauSettings** into const variables before passing them into setAttributes()? 

Or can we simply pass **frauSettings** into setAttributes() and completely remove the need the update this file, should we add additional options in future?

index.html
jorge-ramirez-arredondo: Oh, yup. Definitely just a typo. Should be fixed now.

src/helpers/trackingMethodEnum.js
jorge-ramirez-arredondo: Yeah, it's pretty cool! I don't end up using it too much because usually it's pretty clear when you're not supposed to modify an object anyway, but it adds a bit of extra protection and clarity.

index.html
jorge-ramirez-arredondo: There was no particular reason I left it like that, mostly just wasn't sure yet how much was safe for me to change since I'm still learning how this is all set up. I can update it to pass the whole object in.

jorge-ramirez-arredondo: Okay, I updated this now. Had to use a `...rest` for it though since `page` has special logic.

bamcdole (approved):

index.html
bamcdole: Looks good. Sucks that the CPD component has that one property name discrepancy, but I don't think its a battle worth fighting right now.

PR #253, opened 2021-04-12
Omit attributes with a value of 'false'

bamcdole:
[https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F519816514624&fdp=true?fdp=true](https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F519816514624&fdp=true?fdp=true)

This should fix the issue where our top-level CPD components were parsing input parameters as "true", when they should be "false". In our helper function which attaches input properties to the Lit Component as attributes, we omit properties that would have a value of "false". **This depends on the assumption that any property that would have a value of "false" would be represented by a Boolean property in the corresponding Lit Component.**

I debated between a bunch of different approaches for this fix, and ultimately landed on this one. I think it is the most elegant and concise from a code perspective, and makes the most logical assumptions about how our backend/frontend interface with each other.

This approach makes it the front end's responsibility to treat boolean properties in a special way (which is ultimately a side effect of using LitElement). This is how things worked before, albeit with a more clunky implementation (thanks again @jorge-ramirez-arredondo for refactoring this file!).

Another approach I considered was having the backend omit those properties from being passed to the front end if they would resolve to 'false', but I decided that was a dangerous precedent to set. If we ever wanted to use a different UI framework in place of LitElement, such a pattern would break down immediately.

Totally open to a discussion about different ways to approach this, if anyone has the team has different ideas.


imcdonald-d2l:
> Another approach I considered was having the backend omit those properties from being passed to the front end if they would resolve to 'false', but I decided that was a dangerous precedent to set. If we ever wanted to use a different UI framework in place of LitElement, such a pattern would break down immediately.

I personally would lean in favor of the approach quoted above. I recognize that while this is specific to LitElement, I don't expect this approach to change over the next couple of years. If it does, I don't expect D2L to transition all UI code to the new framework right away. So writing code specific to LitElement I don't see as high risk since it would be years before we need to adjust things.. and that is if CPD is moved to core. The current approach seems "hacky" to me and anyone new to this repo would be very confused on this addition and why we are omitting `false`.

bamcdole:
> > Another approach I considered was having the backend omit those properties from being passed to the front end if they would resolve to 'false', but I decided that was a dangerous precedent to set. If we ever wanted to use a different UI framework in place of LitElement, such a pattern would break down immediately.
> 
> I personally would lean in favor of the approach quoted above. I recognize that while this is specific to LitElement, I don't expect this approach to change over the next couple of years. If it does, I don't expect D2L to transition all UI code to the new framework right away. So writing code specific to LitElement I don't see as high risk since it would be years before we need to adjust things.. and that is if CPD is moved to core. The current approach seems "hacky" to me and anyone new to this repo would be very confused on this addition and why we are omitting `false`.

Would in-line comments documenting the reason for this change help alleviate concerns? The thought of making Lit-specific changes to LMS code feels more "hacky" to me.

For reference, this is what the file looked like previously. You can see that the same sort of approach was taken, albeit on an attribute-by-attribute basis:
https://github.com/Brightspace/continuous-professional-development/blob/8e464a787f5a255fc7226dfaf53b40a9155c1ac0/index.html#L32-L40

If more of the team agrees that the LMS approach is better, I'm happy to spin up a new PR with that change instead.

Syeda-Azim:
Both the change is fine, but I would lean toward the change in the frontend, taking inot account the previous approach that was taken to set the atttributes.

Syeda-Azim (approved):

imcdonald-d2l:
Yes, if we go with this approach a comment explaining why we are adding this logic would be helpful.

jorge-ramirez-arredondo:
I also support the idea of doing the mapping for this in the front-end. In addition to the comment, it might also be worth renaming `setAttributes` to something like `setLitAttributes` or `setLitElAttributes` as a way to help make it clearer that this function is specifically designed to support lit element and it's not just a generic attribute setting function anymore.

index.html
```diff
@@ -17,7 +17,9 @@
 			Object.keys(attributes).forEach((attributeKey) => {
 				let { [attributeKey]: attribute } = attributes;
 
-				if (attribute !== undefined && attribute !== null) {
+				// NOTE: we omit attributes with a value of "false", because LitElement
+				// expects falsey Boolean properties to be 'null', 'undefined', or omitted entirely
+				if (attribute !== undefined && attribute !== null && attribute !== 'false') {
```
jorge-ramirez-arredondo, line 22: > `attribute !== 'false'`
Should `false` be in quotes though? It shouldn't be a string at this point yet, right?

jorge-ramirez-arredondo: Addressed through slack

PR #255, opened 2021-04-13
US125049 Refactoring CPD widget

Syeda-Azim:


src/components/cpd-widget.js
```diff
@@ -148,6 +148,20 @@ class CpdWidget extends BaseMixin(LitElement) {
...
+				denominator: getHoursRounded(progressValues.denominator)
+			};
+		} else {
+			return {
+				numerator: progressValues.numerator,
+				denominator: progressValues.denominator
+			};
+		}
+	}
+
```
Syeda-Azim, line 164: This method is to manipulate the progress values(numerator/denominator) based on the Tracking Method.

src/components/cpd-widget.js
```diff
@@ -148,6 +148,20 @@ class CpdWidget extends BaseMixin(LitElement) {
 		this.selectedView = Number(e.target.value);
 	}
 
+	formatByTrackingMethod(progressValues) {
```
Syeda-Azim, line 151: This method is to manipulate the progress values(numerator/denominator) based on the Tracking Method.

imcdonald-d2l: nit: this could be simplified a bit.
```
const isRecordDuration = this.trackingMethod === trackingMethodEnum.RecordDuration;
return {
  numerator: isRecordDuration ? getHoursRounded(progressValues.numerator) : progressValues.numerator,
  denominator: isRecordDuration ? getHoursRounded(progressValues.denominator) : progressValues.denominator
};
```

src/components/cpd-widget.js
```diff
@@ -199,20 +206,6 @@ class CpdWidget extends BaseMixin(LitElement) {
 		</div>`;
 	}
 
-	renderProgressByType(progressValues, progressString) {
-		return html `
-		<div class="progress-specific">
-			<h4 class="d2l-heading-4">${progressString}</h4>
-			<d2l-meter-radial
```
imcdonald-d2l, line 206: Do you have a screenshot of this change? `renderProgressByType()` uses `<d2l-meter-radial>` but the new function used `renderProgress()` uses `<d2l-meter-circle>`. What is the difference between these two?

Syeda-Azim: My bad, looks like this change needs to be reverted for this difference.

imcdonald-d2l: Yeah, looks like it should be reverted. It also uses different style classes which would affect the UI as well.

imcdonald-d2l (approved):

Syeda-Azim:
When Tracking Method by Record Duration
image

When Tracking Method by Record Duration
image

PR #256, opened 2021-04-13
US125664 CPD - Supporting Record Count to Add/Edit/View Record page

Syeda-Azim:
https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F507759014088


src/components/cpd-add-record.js
```diff
@@ -226,7 +231,9 @@ class AddCpdRecord extends BaseMixin(LitElement) {
 			MethodId: this.shadowRoot.querySelector('#methodSelect').value,
 			IssuedAwardId: this.awardRecord && this.awardRecord.IssuedAwardId,
 			Grade: this.awardRecord && this.awardRecord.Grade,
-			CreditMinutes: getTotalMinutes(this.shadowRoot.querySelector('#creditHours').value, this.shadowRoot.querySelector('#creditMinutes').value),
+			CreditMinutes: this.trackingMethod === trackingMethodEnum.RecordDuration
+				? getTotalMinutes(this.shadowRoot.querySelector('#creditHours').value, this.shadowRoot.querySelector('#creditMinutes').value)
+				: 0,
```
Syeda-Azim, line 236: `CreditMinutes` is a not null column in DB so passing 0 when tracking method is Record Count.

src/components/cpd-add-record.js
```diff
@@ -304,6 +311,7 @@ class AddCpdRecord extends BaseMixin(LitElement) {
 							</li>
 						</ul>
 					</li>
+					${this.trackingMethod === trackingMethodEnum.RecordDuration ? html`
```
imcdonald-d2l, line 314: I would suggest moving the `html` block for the duration condition into its own render function to clean up the code a bit.

src/components/cpd-add-record.js
```diff
@@ -118,6 +122,7 @@ class AddCpdRecord extends BaseMixin(LitElement) {
 		this.methods = [];
 		this.types = [];
 		this.attachments = [];
+		this.trackingMethod = trackingMethodEnum.RecordDuration;
```
imcdonald-d2l, line 125: Why are we manually setting this in the constructor when it should be passed down from `cpd.js`?

src/components/cpd-add-record.js
```diff
@@ -350,6 +337,32 @@ class AddCpdRecord extends BaseMixin(LitElement) {
 		`;
 	}
 
+	renderCreditTime() {
```
imcdonald-d2l, line 340: To be safe we should always return a value in a render function. So in the case where it is record count we should just return ``` html`` ```. 

> Not returning anything (equal to returning undefined in JS) results in bugs where property changes don't trigger a re-render.

For more context see this link: https://github.com/Polymer/lit-element/issues/874#issuecomment-573384599

Had the same issue on my last team.

Syeda-Azim: Interesting! Thanks for sharing.

lweber-d2l (approved):

PR #258, opened 2021-04-14
US127030 - Connect Adjust Targets view to appropriate Record Count API endpoints

jorge-ramirez-arredondo:
https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F521420977592


src/components/cpd-manage-targets.js
```diff
@@ -363,11 +363,10 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
...
 				<td>${this.localize('shared:shortTimeDuration', getHoursAndMinutes(subject.StructuredMinutes + subject.UnstructuredMinutes))}</td>
 			`
-			// TODO: Record-based subject targets will probably use different property names, OR the property names will be different
 			: html `
-				<td>${this.localize('shared:numberOfRecords', getNumberOfRecords(subject.StructuredMinutes))}</td>
-				<td>${this.localize('shared:numberOfRecords', getNumberOfRecords(subject.UnstructuredMinutes))}</td>
-				<td>${this.localize('shared:numberOfRecords', getNumberOfRecords(subject.StructuredMinutes + subject.UnstructuredMinutes))}</td>
+				<td>${this.localize('shared:numberOfRecords', getNumberOfRecords(subject.StructuredRecordCount))}</td>
+				<td>${this.localize('shared:numberOfRecords', getNumberOfRecords(subject.UnstructuredRecordCount))}</td>
+				<td>${this.localize('shared:numberOfRecords', getNumberOfRecords(subject.StructuredRecordCount + subject.UnstructuredRecordCount))}</td>
```
bamcdole, line 369: Does this work with both Tracking Methods?

Syeda-Azim, line 369: `getNumberOfRecords` could be removed as it is redundant here.

jorge-ramirez-arredondo: This part is just for Record Count, the Record Duration part is just above it.

jorge-ramirez-arredondo: `getNumberOfRecords` seems to be wrapping the number in an object before passing it to the localize function (similar to what `getHoursAndMinutes` is doing). Is that not necessary?

Syeda-Azim: The lang term is expecting an object so `getNumberOfRecords` is necessary here

>  "shared:numberOfRecords": "{number} records",

Syeda-Azim (approved):

src/components/cpd-manage-targets.js
bamcdole: Ah, my mistake. I forgot that condition was already in place

bamcdole (approved):

PR #264, opened 2021-04-20
US127227 - CPD - Edit Targets Dialog Should Default to Current Values

Syeda-Azim:
https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F521907501836


src/components/cpd-manage-targets.js
```diff
@@ -237,7 +237,7 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
 						class="numberInput"
 						required type="number"
 						min="0"
-						value="${this.dialogData.StructuredMinutes}"
+						value="${this.dialogData.StructuredRecordCount}"
```
imcdonald-d2l, line 240: Don't we want a condition here to grab either `StructuredMinutes` or `StructuredRecordCount` based on the config set?

data/job_title_target.json
```diff
@@ -1,65 +1,78 @@
 {
-  "TargetId":1,
-  "Type":"JobTitle",
-  "JobTitle":"Sous-Chef",
-  "UserId":175,
-  "StartDate":"0001-12-17T00:00:00.000Z",
-  "Subjects":[
+  "TargetId": 1,
```
imcdonald-d2l, line 2: I don't have context on where `job_title_target.json` and `personal_target.json` are used. Is it just fake data used for testing components?

Syeda-Azim: These json files contains fake data for demo purpose.

src/components/cpd-manage-targets.js
Syeda-Azim: The condition is at the beginning of this method and there is a seperate hmtl to render Record Duration.

imcdonald-d2l (approved):

PR #268, opened 2021-04-21
US126585 Fix to support Show Progress By Subject in My CPD Page

Syeda-Azim:
https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F517481677852


bamcdole (approved):

src/components/progress-box.js
```diff
@@ -72,8 +71,6 @@ class CpdProgressBox extends BaseMixin(LitElement) {
 		this.selectedView = 'overall';
 		this.cpdService = CpdServiceFactory.getCpdService();
 		this.viewUserId = 0;
-		this.trackingMethod = trackingMethodEnum.RecordDuration;
```
IhabAlkabary, line 75: So this line wasn't used:
```
this.trackingMethod = trackingMethodEnum.RecordDuration;
```

IhabAlkabary: I'm guessing it was from before the config were created, and was not cleanup up

Syeda-Azim: `trackingMethod` is passed down from `cpd-my-record.js`, so we wouldn't need to manually set it in constructor.

bamcdole: Normally I would advocate for setting a logical default in the constructor, to handle the case where this component is _not_ passed a trackingMethod.

However, because this property is of Number type, it will default to '0' if not provided via the trackingMethod attribute.

This works in our favour, because our logical default value for the trackingMethodEnum is also '0' 👍

IhabAlkabary (approved):

PR #269, opened 2021-04-22
Converting showProgressBySubject attribute from string to boolean - US126588

imcdonald-d2l:
Fixing issue where property wasn't being set properly since component attribute was being set to a string. Reason is because of default attribute/property binding lit-element provides (good catch @jorge-ramirez-arredondo 🙂 )
image

Fix is to add a custom converter so that we can get a proper boolean value set to the property. See more on lit attribute/property converters [here](https://lit-element.polymer-project.org/guide/properties#conversion-converter).

## gif of it working properly now:
show-progress-config


src/components/cpd-record-report.js
```diff
@@ -64,7 +64,15 @@ class CpdRecordReport extends BaseMixin(LitElement) {
 				type: Number
 			},
 			showProgressBySubject: {
-				type: Boolean
+				type: Boolean,
+				converter: {
+					fromAttribute(value) {
+						if (typeof(value) === 'string') {
```
imcdonald-d2l, line 70: This should always be a string but I took the defensive approach to prevent any string specific actions from blowing up.

lweber-d2l (approved):

src/components/cpd-record-report.js
jorge-ramirez-arredondo: Should this log an error if for some reason we ever get a value that isn't a string? Like if something gets changed in the future and we're getting `undefined` for some reason?

imcdonald-d2l: I don't know of any UI at D2L that logs errors or best practices around that so I would lean on the side of caution before doing that. I will merge this as is and we can revisit this if necessary.

PR #271, opened 2021-04-26
Us127227 - Fix for CPD Edit Target Dialog

Syeda-Azim:
From @bamcdole, who inherited this PR from @Syeda-Azim:

https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F521907501836

This PR addresses the issue where blank values were persisting in the Targets Dialog, as well as validation for the input values for Structured/Unstructured Time/Records.

I also did a bit of refactoring, to try and leave the code in a better state than I found it. I'm hoping this will save us some time, if we need to revisit this component after Noah's redesign efforts.

In lieu of attaching screenshots/videos of the changes in action, I'll aim to do a short demo in post-scrum. We can have a discussion about some of the code/UX decisions made in this PR.


src/components/cpd-manage-targets.js
```diff
@@ -390,7 +381,7 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
 	}
 
 	onDialogDataUpdated(e) {
-		const number = parseInt(e.target.value);
+		const number = parseInt(e.target.value || 0);
```
Syeda-Azim, line 384: Resolving Bad Request when the input field is NaN

src/components/cpd-manage-targets.js
```diff
@@ -48,6 +51,12 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
 			},
 			trackingMethod: {
 				type: Number
+			},
+			invalidTargets: {
+				type: Boolean
+			},
+			expandDialog: {
+				type: Boolean
```
bamcdole, line 59: **invalidTargets** controls whether our Save button is disabled, as a result of validation
**expandDialog** controls whether the height of the dialog should expand to make room for validation tooltips

In an ideal world the d2l-dialog component would know how to resize in response to validation tooltips, but that doesn't seem to be the world we are living in.

src/components/cpd-manage-targets.js
```diff
@@ -65,6 +74,10 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
 				display: grid;
 				grid-template-columns: 1fr 1fr;
 				grid-gap: 25px;
+				margin-right: 1rem;
+			}
+			.dialog-grid.expanded {
+				margin-bottom: 5rem;
```
bamcdole, line 80: The extra right margin is to give validation tooltips enough horizontal space, for fields on the right-hand side of the dialog.

When the dialog is expanded, we are giving validation tooltips on the bottom-most fields enough vertical space.

src/components/cpd-manage-targets.js
```diff
@@ -78,6 +91,16 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
...
 		super();
+
+		this.targetType = Object.freeze({
+			StructuredHours: 'structuredHours',
+			StructuredMinutes: 'structuredMinutes',
+			UnstructuredHours: 'unstructuredHours',
+			UnstructuredMinutes: 'unstructuredMinutes',
+			StructuredRecords: 'structuredRecords',
+			UnstructuredRecords: 'unstructuredRecords'
+		});
```
bamcdole, line 102: I defined a targetType mapping here, to prevent duplicate string identifiers all throughout this component. If any of these need to change in future, it can happen in one place.

src/components/cpd-manage-targets.js
```diff
@@ -105,6 +139,64 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
 			});
 	}
 
+	getRecords(records) {
+		return records || 0;
+	}
```
bamcdole, line 144: A record equivalent of the helper functions in _time-helper.js_. This is only used in this component, so it lives here instead of in a helper file.

src/components/cpd-manage-targets.js
```diff
@@ -105,6 +139,64 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
...
+				break;
+			case this.targetType.StructuredRecords:
+				this.invalidStructuredRecords = targetInvalid;
+				break;
+			case this.targetType.UnstructuredRecords:
+				this.invalidUnstructuredRecords = targetInvalid;
+				break;
+		}
+		this.updateValidity();
+	}
```
bamcdole, line 169: **invalidTargetChange** is an event handler that listens for the _invalid-change_ event from our D2L inputs. This was the best way I could find to keep track of the validation state of our inputs, so that we can enable/disable the Save button _as values are entered_.

src/components/cpd-manage-targets.js
```diff
@@ -105,6 +139,64 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
...
+
+	// d2l-dialog doesn't account for validation tooltips when .resize() is called,
+	// so we have to "make room" for the tooltips on our lower-most properties
+	updateExpandDialog() {
+		this.expandDialog = this.invalidStructuredMinutes
+					|| this.invalidUnstructuredMinutes
+					|| this.invalidStructuredRecords
+					|| this.invalidUnstructuredRecords;
+	}
```
bamcdole, line 178: Only our bottom-most inputs require extra space when a validation tooltip shows up.

I really don't love this solution, but the alternative was a **really** ugly UI state where you had to scroll a tiny window within the dialog box, in order to see the validation tooltip.

src/components/cpd-manage-targets.js
```diff
@@ -105,6 +139,61 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
...
+					|| this.invalidUnstructuredMinutes;
+				break;
+			case trackingMethodEnum.RecordCount:
+				this.invalidTargets = this.invalidStructuredRecords
+					|| this.invalidUnstructuredRecords;
+				break;
+		}
+		this.updateExpandDialog();
+		this.shadowRoot.querySelector('#subject-target-hours-dialog').resize();
+	}
```
bamcdole, line 195: Here we aggregate the "invalid" state of each input into a single property, which will disable our button if it resolves to _true_.

We then update the "expandDialog" property similarly, and force the dialog box to resize.

src/components/cpd-manage-targets.js
```diff
@@ -345,10 +437,9 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
 				<td>
 					${subject.Subject.Name}
 					<d2l-button-icon
-						subject-json="${JSON.stringify(subject)}"
 						icon="tier1:edit"
 						text="${this.localize('manageTargets:editSubjectTarget', {subject: subject.Subject.Name})}"
-						@click="${this.openSubjectTargetDialog}">
+						@click="${this.openSubjectTargetDialogHandler(subject)}">
```
bamcdole, line 442: I really didn't like how this component was passing the "subject" object to the click handler here.

The old way involved stringify-ing the object and injecting it into an attribute, which was then fetched from the DOM via the Click Event's attributes pointer, here:
https://github.com/Brightspace/continuous-professional-development/pull/271/files#diff-c9668013e3fb5011410bdec3ad40aed6f0d9c09198c1a0f9f33d95c6a4278103L424-L425

src/components/cpd-manage-targets.js
```diff
@@ -420,17 +518,29 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
...
+					break;
+				case trackingMethodEnum.RecordCount:
+					this.dialogData = {
+						StructuredRecordCount: subject.StructuredRecordCount,
+						UnstructuredRecordCount: subject.UnstructuredRecordCount
+					};
+					break;
+			}
+			this.dialogData.SubjectId = subject.Subject.Id;
+			this.shadowRoot.querySelector('#subject-target-hours-dialog').open();
```
bamcdole, line 542: This is my proposed alternative for passing 'subject' to the click handler, mentioned [here](https://github.com/Brightspace/continuous-professional-development/pull/271/files#r621536205).

bamcdole: Another notable change here is how we are constructing the **dialogData** property, which is the data that is manipulated by the Targets Dialog.

For duration-based tracking, the previous implementation only kept track of _minutes_. This created a disjoint between the data and the UX, which necessitated a jarring workflow where editing _minutes_ would affect the value of _hours_, and vice versa. 

Instead, I am take the backend data (which tracks by minutes only), and transforming it into an object which tracks _minutes_ and _hours_ separately. This allows the UX to work as expected.

When the dialog is saved, we transform the data back into just _minutes_ before calling the save API.

src/components/cpd-manage-targets.js
```diff
@@ -449,10 +559,19 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
...
-		} else {
-			await this.cpdService.updateTargetRecordCount(this.jobTitle, this.dialogData);
+
+		switch (this.trackingMethod) {
+			case trackingMethodEnum.RecordDuration:
+				// If we are tracking my duration, switch back to only minutes
+				this.dialogData.StructuredMinutes = getTotalMinutes(this.dialogData.StructuredHours, this.dialogData.StructuredMinutes);
+				this.dialogData.UnstructuredMinutes = getTotalMinutes(this.dialogData.UnstructuredHours, this.dialogData.UnstructuredMinutes);
+				delete this.dialogData.StructuredHours;
+				delete this.dialogData.UnstructuredHours;
```
bamcdole, line 569: This is where we transform our **dialogData** model back into an object that only holds _minutes_, which is what the API expects.

src/components/cpd-manage-targets.js
Syeda-Azim: Don't we have to define `targetType` as a Property of 'Object' type? I understand it's working fine without it but not sure of the intention behind it.

src/components/cpd-manage-targets.js
```diff
@@ -170,57 +259,63 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
...
 				<div id="structured">
-					<label for="structuredHours">${this.localize('manageTargets:hours')}</label>
-					<d2l-input-text
-						@blur="${this.onDialogDataUpdated}"
-						id="structuredHours"
+					<label for=${this.targetType.StructuredHours}>${this.localize('manageTargets:hours')}</label>
+					<d2l-input-number
+						@change="${this.onDialogDataUpdated}"
+						id=${this.targetType.StructuredHours}
 						class="numberInput"
```
Syeda-Azim, line 266: We can get rid of `class`and `type` attributes from <d2l-input-number>

Syeda-Azim: Also, `label` tags could be replaced with the label attribute in `d2l-input-number`, unless it is changing the UI drastically.

bamcdole (approved):

PR #278, opened 2021-04-29
US127708 - Validation on add/edit records page

bamcdole:
https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F602284549980&fdp=true?fdp=true

These changes aim to bring our validation on the Add/Edit Record page up to par with the recent changes to the Adjust Targets dialog (https://github.com/Brightspace/continuous-professional-development/pull/271). 

In addition to the issues outlined in the Rally case, there is one additional fix. The previous implementation would not remove red outlines from blank inputs until "Save" was re-pressed. Now, the inputs will validate input _as it is entered_, which is consistent with the rest of Brightspace.

The main change is switching the page to leverage built-in validation provided by our Core input components, instead of applying 'aria-invalid' to the input elements manually.

I also did some refactoring to make this file more maintainable in future (reduce number of hardcoded string IDs, use the dedicated d2l-input-number component, etc.)


src/components/cpd-add-record.js
```diff
@@ -346,12 +365,30 @@ class AddCpdRecord extends BaseMixin(LitElement) {
...
+									<d2l-input-number
+										id="${this.inputId.CreditHours}"
+										min="0"
+										value="${this.record && !isNaN(this.record.CreditMinutes) ? getHours(this.record.CreditMinutes) : 'blank'}"
+										input-width="200px"
+										max-fraction-digits="0"
+										required
+										placeholder=${this.localize('addRecord:enterCreditHours')}>
+									</d2l-input-number>
+					</d2l-input-number>
```
imcdonald-d2l, line 378: extra unnecessary closing tag for `d2l-input-number`

imcdonald-d2l (approved):

PR #280, opened 2021-04-30
US127227 - Edit Target dialog - "Cancel" bug

bamcdole:
Cleaned up some HTML, fixed a bug where Cancelling out of the Edit Target dialog would prevent the input components from detecting a change in 'value' when the dialog is reopened.

https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F521907501836&fdp=true?fdp=true


src/components/cpd-manage-targets.js
```diff
@@ -122,6 +122,11 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
 		this.expandDialog = false;
 	}
 
+	cancel() {
+		// For some reason, this forces the d2l-input-number components to detect the change in 'value'
+		this.subjectTargets = Object.assign({}, this.subjectTargets);
+	}
```
bamcdole, line 128: This is the fix - almost everything else is code cleanup

bamcdole:
This fix should now work regardless of how the dialog is closed.

Syeda-Azim (approved):

PR #281, opened 2021-05-04
Fix dialog issue on admin page - US127790

imcdonald-d2l:
### Issue
CPD iframe dynamically resizes based on the contents within it (setup in index.html [here](https://github.com/Brightspace/continuous-professional-development/blob/master/index.html#L38)). When no questions/subjects/methods exist the size of the iframe shrinks and does not leave enough height for the dialog. Since the dynamic resizer is not smart enough to consider dialogs I had to set a min-height on the tabs component to give enough room for the dialog. 

### Before
dialog-before

### After
dialog-after

Rally: https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F602781623012


IhabAlkabary:
I would expand it a tiny bit more, since it's a bit tight under the buttons

IhabAlkabary:
or maybe get Noah's feedback on this if we need to increase the min height just a bit more

imcdonald-d2l:
@IhabAlkabary buttons have always been like that in the dialog - I agree that we need more padding under the buttons though. I didn't adjust that in this PR since I was going to leave that to the UX overhaul stories.

IhabAlkabary:
> @IhabAlkabary buttons have always been like that in the dialog - I agree that we need more padding under the buttons though. I didn't adjust that in this PR since I was going to leave that to the UX overhaul stories.

If this is how it always looks, then yeah, let's leave it for the UX improvement feature

IhabAlkabary (approved):

PR #283, opened 2021-05-10
Switch tests to use Karma

imcdonald-d2l:
Up-to-date PR here: https://github.com/Brightspace/continuous-professional-development/pull/286

Will be closed shortly.


karma.conf.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-env node */
```
imcdonald-d2l, line 1: Using template found here: https://github.com/BrightspaceUI/create/blob/master/src/generators/test-unit/templates/static/karma.conf.js

PR #284, opened 2021-05-11
Refactor CPD build logic - both dev and prod - US128020

imcdonald-d2l:
Refactoring CPD build logic so it aligns with more modern build practices as well as supporting more up-to-date BSI components where `polymer` commands fail to support things. (see [slack thread](https://d2l.slack.com/archives/C0PHG3QB0/p1620658304263700) for example)

In short, we are now using `rollup` for production builds and `es-dev-server` for local builds. Both of these tools modernize our build approach and give us greater flexibility for the future. Comments in-line will provide further explanation as to how it was implemented and documentation for further clarification.

Rally ticket: https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F603926361524


.github/workflows/ci.yml
```diff
@@ -19,9 +19,10 @@ jobs:
...
-        env:
-          SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY_DESIRE2LEARN }}
-          SAUCE_USERNAME: Desire2Learn
+        run: npm run build:prod
+      # Unit tests to be fixed in a follup up PR
+      # - name: Unit Tests (SauceLabs)
+      #   run: npm run test:sauce
+      #   env:
+      #     SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY_DESIRE2LEARN }}
+      #     SAUCE_USERNAME: Desire2Learn
```
imcdonald-d2l, line 28: Switching to Karma tests in a follow-up PR. 

Draft one started here: https://github.com/Brightspace/continuous-professional-development/pull/283

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules/
 build/
 .vs/
+appconfig.json
```
imcdonald-d2l, line 4: `appconfig.json` gets added by `npm run start:dev` at the root level from `frau-appconfig-builder --dist ./`. Adding it here so it doesn't get committed. When we override the FRA it will now be at `http://localhost:3000/appconfig.json`

es-dev-server-dev.config.js
```diff
@@ -0,0 +1,14 @@
+const cors = require('@koa/cors');
+
+module.exports = {
```
imcdonald-d2l, line 3: For more context on what these values mean you can look at the documentation [here](https://github.com/open-wc/es-dev-server#code-transformation).

package.json
```diff
@@ -9,16 +9,17 @@
...
       "loader": "iframe"
     },
     "frauLocalAppResolver": {
       "appClass": "urn:d2l:fra:class:continuous-professional-development",
-      "dist": "./build/dev",
-      "hostname": "localhost"
+      "dist": "./",
+      "hostname": "localhost",
+      "port": "3000",
+      "baseRoute": ""
```
imcdonald-d2l, line 22: `baseRoute` is a new addition to `frauLocalAppResolver`. For context on what it does you can refer to the following PR: https://github.com/Brightspace/frau-local-appresolver/pull/35

package.json
```diff
@@ -29,42 +30,36 @@
     }
   },
   "scripts": {
```
imcdonald-d2l, line 32: removing all pre-existing build commands that are now irrelevant as well as any testing commands that will be replaced in a follow-up PR that implements Karma for testing. Draft testing PR can be found here: https://github.com/Brightspace/continuous-professional-development/pull/283

package.json
```diff
@@ -29,42 +30,36 @@
...
     "publish:dev:cdn": "frau-appconfig-builder && frau-publisher",
-    "resolver": "frau-local-appresolver",
-    "test": "npm run test:local && npm run test:mocha && npm run lint",
-    "test:local": "polymer test --skip-plugin sauce",
-    "test:sauce": "polymer test --skip-plugin local",
-    "test:travis": "npm run test:mocha && polymer test --config-file wct.conf-travis.json",
-    "test:mocha": "mocha --require esm",
-    "watch:js": "onchange -d 2000 \"src\\**\" \"index.html\" -- npm run build:local",
-    "watch": "npm-run-all build:local --parallel resolver watch:js"
+    "build:prod": "rimraf build && rollup -c rollup.config.js",
```
imcdonald-d2l, line 35: For prod builds we are using `rollup`. This allows us to step away from `polymer build` commands and use more modern tooling. Configuration for `rollup` command can be found in `rollup.config.js`.

package.json
```diff
@@ -29,42 +30,36 @@
...
+    "start:dev": "frau-appconfig-builder --dist ./ && concurrently --kill-others \"es-dev-server --app-index index.html --config es-dev-server-dev.config.js --watch\""
```
imcdonald-d2l, line 36: For dev local builds we are using `es-dev-server`. Essentially `es-dev-server` is a web server for developing without having to bundle anything. It hosts files directly from your file system. Also gives us hot-reloading 🎉 

For more information on `es-dev-server` you can refer to the following links:

- https://github.com/open-wc/es-dev-server
- https://dev.to/open-wc/developing-without-a-build-2-es-dev-server-1cf5

rollup.config.js
```diff
@@ -0,0 +1,34 @@
+import copy from 'rollup-plugin-copy';
+import { createSpaConfig } from '@open-wc/building-rollup';
+import merge from 'deepmerge';
+
+const config = createSpaConfig({
```
imcdonald-d2l, line 5: For more information on this config please refer to links referred to in `package.json`.

rollup.config.js
```diff
@@ -0,0 +1,34 @@
...
+	'node_modules/d2l-html-editor/langs/**',
+	'node_modules/d2l-html-editor/d2l_lang_plugin/**',
+	'node_modules/d2l-html-editor/**/*.css'
+];
+
+export default merge(config, {
+	input: './index.html',
+	plugins: [
+		copy({
+			// used to copy over assests such as images or json files
```
imcdonald-d2l, line 23: Copying over assets that `rollup` doesn't automatically catch. I their own words: 
>To copy over assets, such as images, css or json files, we recommend using rollup-plugin-copy

Documentation here: https://open-wc.org/docs/building/rollup/#copying-assets

src/components/cpd-loader.js
```diff
@@ -1,4 +1,4 @@
-import './cpd';
-import './cpd-admin';
-import './cpd-widget';
-import './cpd-record-report';
+import './cpd.js';
+import './cpd-admin.js';
+import './cpd-widget.js';
+import './cpd-record-report.js';
```
imcdonald-d2l, line 4: Adding `.js` to imports so that `rollup` can fetch and bundle all dependencies as expected.

demo/admin.html
```diff
@@ -2,10 +2,10 @@
 <html lang="en">
 	<head>
 		<link rel="stylesheet" href="../node_modules/@brightspace-ui/core/components/demo/styles.css" type="text/css">
-		<script src="../node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
```
dlockhart, line 5: I think you'll want to keep the polyfill here for now, since `demo-page.js` won't work without the polyfill loaded first. Here's our sample demo file for reference from our NPM initializer:
https://github.com/BrightspaceUI/create/blob/master/src/generators/demo/templates/index.html

es-dev-server-dev.config.js
dlockhart: I'm guessing you needed the CORS middleware? Otherwise I think you could accomplish this just with command line args.

package.json
```diff
@@ -29,42 +30,36 @@
...
+    "es-dev-server": "^2.1.0",
     "eslint": "^6",
     "eslint-config-brightspace": "^0.6",
     "eslint-plugin-html": "^6",
     "eslint-plugin-lit": "^1",
     "eslint-plugin-sort-class-members": "^1",
     "esm": "^3.2.25",
-    "frau-appconfig-builder": "^1.0.0",
+    "frau-appconfig-builder": "^1",
     "frau-ci": "^1",
```
dlockhart, line 56: Unrelated to this PR: you shouldn't need `frau-ci` anymore since you're deploying using the incremental release GitHub action.

package.json
dlockhart: Do you need concurrently here? Seems like it's just running a single command.

package.json
dlockhart: Might be nice if you had a `start:prod` so that you could test the prod build locally. You may then just want to rename `start:dev` to `start` which is a standard NPM command and can be run with `npm start` instead of `npm run start`. 🤷

demo/admin.html
imcdonald-d2l: Sounds good, I'll revert these changes 👍🏻

es-dev-server-dev.config.js
imcdonald-d2l: Good catch - forgot to test without this. It actually isn't needed. I'll remove it and see if we can get rid of the config file and just use command line. 👍🏻

package.json
imcdonald-d2l: Good catch - no, I don't believe it is necessary.

demo/admin.html
dlockhart: Double-check that you're not getting 404s now, I've seen that before. It may need to be `/node_modules/` instead of `../node_modules/`.

package.json
imcdonald-d2l: Good point - I'll add a `start:prod` command for testing 👍🏻

imcdonald-d2l: Added the `start:prod` command for testing!

es-dev-server-dev.config.js
imcdonald-d2l: Config file has been removed and just using command line with args now

demo/admin.html
imcdonald-d2l: I doubled checked and it was working fine already. But I also noticed it worked just fine without `/node_modules/` so I just removed it all together.

dlockhart (approved):
Looks good! Really impressive you were able to get all this working -- hoping this can serve as a blueprint for others to get iFRAs migrated over to modern build tooling. 🥇

package.json
```diff
@@ -9,16 +9,17 @@
   },
   "config": {
     "frauAppConfigBuilder": {
-      "appFile": "./index.html",
-      "distLocal": "./build/dev",
+      "appFile": "build/prod/index.html",
```
imcdonald-d2l, line 12: @dlockhart this part I'm a little unsure of. Should `appFile` stay as `index.html` for when we run `publish:dev:cdn`? I had to adjust this for `npm run start:prod` but could alternatively just specify `--appfile` in line for that command. Just don't know if `frau-publisher` expects `appFile` a certain way. 🤔

dlockhart: Yeah not totally sure, you may have to experiment when this actually publishes and then tweak.

lweber-d2l (approved):
Thanks for the helpful explanatory comments.  I checked out the branch and verified
`npm start` and that the following pages loaded:
http://localhost:3000/demo/index.html
http://localhost:3000/demo/admin.html
http://localhost:3000/demo/progress.html
http://localhost:3000/demo/report.html
I'm not an expert in the JS build area so it's good that Dave approved this too 😄 
Nice work!

package.json
imcdonald-d2l: For those reviewing this PR, this needed to be updated to just `index.html`. 
PR can be found here: https://github.com/Brightspace/continuous-professional-development/pull/285

PR #287, opened 2021-05-27
removing unused table references

dlockhart:
As far as I can tell, nothing from `d2l-table` is actually being used in this repo. Maybe it was at one time? Regardless, I'm going through and switching old table references to the new Lit one and this seemed like an easy target. :)


dlockhart:
Dug a little deeper -- looks like table was originally [used 2+ years ago](https://github.com/Brightspace/continuous-professional-development/commit/df3bae522deb2fdfca9889a182564f98f8160bc0) but has since been removed (history i s pretty hard to track due to moved files). FYI @BraydonKains

imcdonald-d2l (approved):

PR #289, opened 2021-05-31
Record Dialog - us127855

imcdonald-d2l:
New record dialog for create/edit/read-only records. Comments will be added for better clarification.

**Note**: a lot of the new dialog component is copied from the deleted `cpd-add-record` and `cpd-read-only-record`. I cleaned up some code and styles when combining the components but there is likely still a lot of cleanup that can be done. So if you see any issues or areas for cleanup, let me know! 

### Create Record
cpd-create-dialog

### Edit Records
cpd-edit-dialog

### Read Only Records
Ignore gross styling - this should be cleaned up later once Josh and Noah review. I decided to disable the save button in read only view  - we might want to remove it all together at a later date if that's what Josh and Noah want.
cpd-read-only-record

### New record via route
cpd-new-record-route


src/components/cpd-my-records.js
```diff
@@ -235,27 +277,25 @@ class MyCpdRecords extends BaseMixin(LitElement) {
...
 		}
+	}
 
+	openDialog(newRecord, recordId = null, isReadonly = false) {
+		// openning dialog without routing
+		this.recordId = recordId;
+		this.readOnlyRecord = isReadonly;
+		this.editDialog = !newRecord;
+		this.openDialogWithoutRouting = true;
+		this.closeDialogWithoutRouting = true;
```
imcdonald-d2l, line 298: Unfortunately we need both these properties when opening a new dialog. `openDialogWithoutRouting` opens/closes the dialog. And `closeDialogWithoutRouting` ensures we don't re-route on close. Hopefully this can be cleaned up with internal app routing.

src/components/cpd-record-dialog.js
```diff
@@ -0,0 +1,553 @@
...
+	}
+
+	async firstUpdated() {
+		// Give the browser a chance to paint
+		// we cannot open the dialog until browser has rendered the components
+		// or we will get a focus error and cannot listen to events
+		await new Promise((r) => setTimeout(r, 0));
+		if (this.showRecordDialog) {
+			this.openRecordDialog = true;
+		}
```
imcdonald-d2l, line 195: Normally, the browser page would already be rendered before opening a dialog in normal use cases but since we want to open it immediately we have to wait until after first paint before opening it. If not you'll get gross console errors and things won't function properly.

Using [Lit Element's suggestion](https://lit-element.polymer-project.org/guide/events#add-event-listeners-after-first-paint) for adding logic after first paint.

src/components/cpd-record-dialog.js
```diff
@@ -0,0 +1,553 @@
...
+					<label>${this.localize('addRecord:addEvidence')}</label>
+					<d2l-attachments .attachmentsList="${this.attachments}" @d2l-attachments-list-updated="${this.attachmentsUpdated}"></d2l-attachments>
+				</div>
+			</div>
+			${this.questions.map((q) => this.renderQuestion(q))}
+		`;
+	}
+
+	renderReadOnlyRecord() {
+		return html`
```
imcdonald-d2l, line 435: combining `read-only-record` into the dialog as well because it uses a lot of the same styles/logic as `cpd-add-records`. It makes the file larger but reduces copied code and need for updating two components for one UI style change.

src/components/cpd-record-dialog.js
```diff
@@ -0,0 +1,553 @@
...
+				CreditMinutes: this.awardRecord.CreditMinutes,
+				DateCompleted: dateParamString(this.awardRecord.IssuedDate)
+			};
+		}
+	}
+
+	updated(changedProperties) {
+		if (changedProperties.has('recordId')) {
+			this.fetchRecord();
+		}
```
imcdonald-d2l, line 185: This will update the form with record details when editing various records.

src/components/cpd-my-records.js
```diff
@@ -164,12 +194,24 @@ class MyCpdRecords extends BaseMixin(LitElement) {
...
+		window.removeEventListener('record-dialog-closed', (e) => this.dialogClosedAction(e));
 		super.disconnectedCallback();
 	}
 
+	dialogClosedAction(e) {
+		this.openDialogWithoutRouting = false;
+		this.editDialog = false;
+		this.recordId = null;
+		if (e.detail.type === SAVE_ACTION) {
+			this.fetchRecords();
```
imcdonald-d2l, line 211: If saving a new or saving an edit on an existing record - fetch the records again to get all updates.

src/components/cpd-record-dialog.js
jorge-ramirez-arredondo: Is this something that could be fixed on the component side? Do you think it's worth opening a github issue for the dialog?

src/components/cpd-my-records.js
```diff
@@ -235,27 +277,25 @@ class MyCpdRecords extends BaseMixin(LitElement) {
...
-					recordId: e.target.getAttribute('record-id'),
-					viewUserId: this.viewUserId
-				}
-			);
+			this.openDialog(false, recordId, true);
 		}
+	}
 
+	openDialog(newRecord, recordId = null, isReadonly = false) {
+		// openning dialog without routing
```
lweber-d2l, line 293: Very minor - spelling: `openning` => `opening`

src/components/cpd-record-dialog.js
```diff
@@ -0,0 +1,561 @@
...
+		if (this.record.Attachments) {
+			removedAttachments = this.record.Attachments.Files
+				.filter(oldFile => !this.attachments.includes(oldFile))
+				.map(f => f.id);
+		}
+		this.cpdService.updateRecord(this.recordId, record, newAttachments, removedAttachments)
+			.then(() => {
+				this.closeDialog(SAVE_ACTION);
+				this.saving = false;
+			});
```
jorge-ramirez-arredondo, line 355: Since you made `saveUpdatedRecord` async now, should you be awaiting on the `this.cpdService.updateRecord(...)` call instead of using `.then(...)`?

src/components/cpd-record-dialog.js
imcdonald-d2l: I'm not sure if it's worth a GitHub issue for this since this particular case since it is an edge case for dialogs at D2L. Don't think there are many places where people would want a dialog opened on a new route.

src/components/cpd-record-dialog.js
imcdonald-d2l: Unfortunately `this.cpdService.updateRecord()` is not an `async` function so we can't `await` on it 🙁

lweber-d2l (approved):
Looks good 👍

src/components/cpd-record-dialog.js
jorge-ramirez-arredondo: Hmmm, yeah, it probably wouldn't be a super common scenario, but if we're going to be doing front-end routing more often, tying a route to a dialog will probably pop up here and there. I'll leave it up to you if you think it's worth it though. I just thought I would bring it up as an option.

src/components/cpd-record-dialog.js
jorge-ramirez-arredondo: Huh, why does it have a `.then()` then? It's not returning a promise?

imcdonald-d2l: Ah... you are correct. It is returning a promise so I can `await` it. I have updated the code with your suggestion 🙂

src/components/cpd-record-dialog.js
imcdonald-d2l: Yeah, I agree. Certainly good to bring this up as a GitHub issue if this a re-occurring issue for us.

jorge-ramirez-arredondo (approved):

PR #294, opened 2021-06-18
New html editor - us127857

imcdonald-d2l:
Adding new html editor to CPD as well as adjusting margin on `d2l-tabs` component to remove CPD jitter. 

**NOTE:** the dropdowns going behind following editors is a known issue and will hopefully be resolved by the appropriate team soon (🤞🏻 ). See slack thread [here](https://d2l.slack.com/archives/C0PHG3QB0/p1623937132292400).  If we're not updating production CPD for a bit I don't see an issue merging this. 

### New HTML editor
new-html-editor


src/components/cpd.js
```diff
@@ -51,6 +51,7 @@ class Cpd extends BaseMixin(LitElement) {
 			}
 			d2l-tabs {
 				min-height: 800px;
+				margin: 0;
```
imcdonald-d2l, line 54: This fixes the jitter in the CPD app. I used process of elimination to determine the component that was causing us issues and it turned out to be the `d2l-tabs` component. Why this fixes the issue... well I'm still not exactly sure but it seems to fix the issue so I'll take it. 😆  🤷🏻‍♂️ 

My guess is that it seems to be adding and removing some margin every time an action is made within the tabs container. 

### With Margin
Take note of the resize events in the console every time an action is made on the page. 
all-the-resize-events


### Without Margin
Note that there are no longer any unnecessary resize events when performing actions on the page.
no-resize-events

.github/workflows/ci.yml
```diff
@@ -14,6 +14,8 @@ jobs:
         uses: Brightspace/third-party-actions@actions/checkout
       - name: Setup Node
         uses: Brightspace/third-party-actions@actions/setup-node
+        with:
+          node-version: '14.x'
```
bamcdole, line 18: Is this a requirement for the new HTML editor?

bamcdole (approved):

.github/workflows/ci.yml
imcdonald-d2l: no - was getting a really weird `TypeError: quasi.value.raw.matchAll is not a function` linting error in CI. Updating to node LTS fixed that issue.

PR #297, opened 2021-06-21
Fix attachment dialog issues - de43823

imcdonald-d2l:
Prevent both dialogs from being closed when close action is emitted from the attachments dialog.

### Working dialog
record-attachments

https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F600942237231


src/components/cpd-record-dialog.js
```diff
@@ -208,7 +210,12 @@ class RecordDialog extends BaseMixin(LitElement) {
 		this.attachments = event.detail.attachmentsList;
 	}
 
-	async cancelDialogAction() {
+	async cancelDialogEvent(e) {
+		if (e.path[0].id != DIALOG_ID) return;
```
imcdonald-d2l, line 214: Both the attachments and record dialog emit a `d2l-dialog-close` event on cancel/save actions. Both dialogs will catch these events so we have to check the origin of the event before actually closing the record dialog.

src/components/attachments.js
```diff
@@ -28,6 +28,9 @@ class Attachments extends BaseMixin(LitElement) {
 
 	static get styles() {
 		return css`
+			.d2l-add-attachments-btn {
+				padding-bottom: 8px;
```
imcdonald-d2l, line 32: Just adding padding to attachments button to prevent style overlap with attachments list.

bamcdole (approved):

PR #298, opened 2021-06-22
Include tinymce packages in rollup - DE44115

imcdonald-d2l:
I'm not 100% confident this will solve the issue since local testing of the production build doesn't seem to be working with this change. But since this change won't break anything it will be safe to merge and test to see if it fixed anything.

https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F602335203685&fdp=true


bamcdole (approved):

imcdonald-d2l:
Update: did not solve the issue. 😢

PR #300, opened 2021-06-23
Read only record bugs - DE43825, DE43837

imcdonald-d2l:
Fixes:
- Credit hours and evidence field were duplicated in read only view (html markup was duplicated - just needed to remove duplicated code)
- Don't show record duration field when track by record count is enabled
- Small styling fixes

### Working read-only view
fix-readonly

https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F600942861063
https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F601049965885


src/components/cpd-record-dialog.js
```diff
@@ -416,6 +417,8 @@ class RecordDialog extends BaseMixin(LitElement) {
 	}
 
 	renderReadOnlyRecord() {
+		if (!this.record) return html``;
```
imcdonald-d2l, line 420: Added this to prevent `this.record.name` property undefined error that was appearing in the console.

src/components/cpd-record-dialog.js
```diff
@@ -499,8 +501,10 @@ class RecordDialog extends BaseMixin(LitElement) {
 
 	renderQuestion(question) {
 		return html`
+			<label for="${`answerText_${question.Id}`}" class="d2l-label-text">${question.QuestionText}</label>
```
imcdonald-d2l, line 504: Adding explicit label to question instead of using html editor label property since it wasn't properly displaying the label in read-only view. Definitely a bug on the html editor side but adding this to avoid it all-together.

src/components/cpd-record-dialog.js
```diff
@@ -539,7 +543,7 @@ class RecordDialog extends BaseMixin(LitElement) {
 		if (attachments) {
 			return html`<d2l-attachments readonly .attachmentsList="${attachments.Files}"></d2l-attachments>`;
 		}
-		return this.localize('readOnlyRecord:noEvidence');
+		return html`<p>${this.localize('readOnlyRecord:noEvidence')}</p>`;
```
imcdonald-d2l, line 546: Adding `<p>` tag to get margin around text.

src/components/cpd-record-dialog.js
imcdonald-d2l: readonly-removing-label

imcdonald-d2l: Created github issue here: https://github.com/BrightspaceUI/htmleditor/issues/271

src/components/cpd-record-dialog.js
jorge-ramirez-arredondo: Isn't this a redundant check? It looks like `renderReadOnlyRecord` doesn't get called unless `this.readOnly && this.record` is true to begin with. How are we getting in here if `this.record` doesn't exist?

imcdonald-d2l: Ah good point - just realized I was getting this error because I was forcefully calling `renderReadOnlyRecord` when testing these changes so I didn't have to setup teams to get read-only records. I'll remove this change.

jorge-ramirez-arredondo (approved):

PR #302, opened 2021-06-23
DE44071 Fix routing issues when viewing team records

jorge-ramirez-arredondo:
https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F602132571173&fdp=true?fdp=true


src/components/cpd.js
```diff
@@ -108,12 +108,17 @@ class Cpd extends BaseMixin(LitElement) {
...
+
 		return html`
 			<d2l-cpd-my-records
-				viewuserid="${this.pageData.viewUserId}"
 				trackingMethod="${this.trackingMethod}"
-				?showProgressBySubject="${this.showProgressBySubject}">
-			</d2l-cpd-my-records>
+				?showProgressBySubject="${this.showProgressBySubject}"
+				viewUserId="${ifDefined(viewUserId)}"
+				?readOnlyRecord="${viewUserId !== null && viewUserId !== undefined}"
```
imcdonald-d2l, line 118: nit: you should be able to just do `!!viewUserId`
https://dev.to/sanchithasr/what-is-the-double-bang-operator-in-javascript-4i3h

imcdonald-d2l (approved):

src/components/cpd.js
jorge-ramirez-arredondo: You actually wouldn't need the double bang at all in this case since `?` in front of `?readOnlyRecord` should already evaluate truthy/falsey values to true/false:
https://lit-element.polymer-project.org/guide/templates#bind-to-a-boolean-attribute

That being said, the reason I had it that way to begin with was just me being overly defensive. When I originally wrote that, I wasn't sure if `viewUserId` was a number or a string, so I decided to guard against the possibility of it being the number `0`.

jorge-ramirez-arredondo:
CPD `V2.5.11`

src/components/cpd.js
imcdonald-d2l: Ah interesting, didn't know the `?` evaluates objects like that. Good to know 👍🏻 
But yes, doesn't hurt being defensive.

PR #309, opened 2021-07-07
DE44147 - CIPD Filter by Date showing records outside of selected range by one day

Syeda-Azim:
https://rally1.rallydev.com/#/431372673576d/dashboard?detail=%2Fdefect%2F602461243651


src/helpers/time-helper.js
```diff
@@ -7,13 +7,8 @@ export function formatForDatePicker(date) {
 	return dayjs(date).format('YYYY-MM-DD');
 }
 
-export function dateParamString(dateStr, end = false) {
+export function dateParamString(dateStr) {
 	let date = dayjs(dateStr);
-
-	if (end) {
-		date = date.endOf('day');
-	}
```
Syeda-Azim, line 15: The time of End Date was modified here to end of day time which then converted to next day in local time.

test/helpers/time-helper.tests.js
```diff
@@ -14,11 +14,11 @@ describe('TimeHelpers', () => {
 		});
 		it('should produce end of day if end', () => {
```
lweber-d2l, line 15: Very minor: since `dateParamString()` no longer takes `end` as a parameter, the expectations could be updated for clarity.  Also, both tests are now checking for the same thing, so one of them could be removed.

Syeda-Azim:
All the Records in DB are 4 hours ahead of the day due to datetime conversion. So to include records in EndDate Range we will have to extend the EndDate filter to EOD.

lweber-d2l:
> All the Records in DB are 4 hours ahead of the day due to datetime conversion. So to include records in EndDate Range we will have to extend the EndDate filter to EOD.

Hmmm... I recall we had to deal with something similar for ADS Scheduler, front end was in local time, but DB and back end were in UTC... fun times. 😄

PR #310, opened 2021-07-07
Change title sync to not be enabled for the widget

jorge-ramirez-arredondo:
[DE43560](https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F604646852848&fdp=true?fdp=true)

image


lweber-d2l (approved):
Nice, clean solution. 👍 Looks good as far as I can tell, with the caveat that I'm not very familiar with `ifrau`.

jorge-ramirez-arredondo:
`v2.5.17`

PR #315, opened 2021-07-14
DE44310 Remove attachments from list

julianpettit:
Adds removal logic to the add attachments dialog list

https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F603570808415


src/components/attachments.js
```diff
@@ -75,6 +75,12 @@ class Attachments extends BaseMixin(LitElement) {
 		this.fireAttachmentListUpdated(oldVal);
 	}
 
+	updateCurrentAttachments(evt) {
+		const oldVal = [...this.currentAttachments];
+		this.currentAttachments = [...evt.detail.attachmentsList];
+		this.fireAttachmentListUpdated(oldVal);
```
jorge-ramirez-arredondo, line 81: I didn't notice this until now, but `this.fireAttachmentListUpdated(...)` here is firing an event, right? What is this event being used for? Does it make sense for it to fire when making changes to the `currentAttachments` list?

bamcdole (approved):

src/components/attachments.js
bamcdole: Good questions, I am also curious.

I assumed the purpose of the event was for notifying other components that _this_ component's currentAttachments had changed, but I don't actually know what the event is used for currently.

julianpettit: Right now it doesn't look like `internal-attachments-list-updated` is used anywhere, either for `currentAttachments` or `attachmentsList`

jorge-ramirez-arredondo: `this.fireAttachmentListUpdated(...)` is firing an event called `d2l-attachments-list-updated` though, not `internal-attachments-list-updated`.

jorge-ramirez-arredondo: Based on what I see just in this file, it looks like `d2l-attachments-list-updated` is fired any time `attachmentsList` is updated. I'm not sure it would make sense to also trigger this event when `currentAttachments` is updated.

julianpettit: Good point, in that case probably better to leave it off the handler

jorge-ramirez-arredondo (approved):

PR #316, opened 2021-07-22
US129946 fix file list formatting

julianpettit:
- Removed the unused grid layout for the attachment list component
- Removed the file size line break
- Added overflow ellipses to large file names (previously large files without breaks pushed the X button off of the screen)

image
https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F603967669735


TomMichielsD2L (approved):
LGTM.
Not exactly a new problem, but the x button looks vertically off center. Should that be fixed too?

lweber-d2l:
> LGTM.
> Not exactly a new problem, but the x button looks vertically off center. Should that be fixed too?

Yeah, I see it too - the x looks positioned a little low on each line.  It would be ok to fix that here, or in a separate PR - either is good.

lweber-d2l (approved):

PR #317, opened 2021-07-22
DE44147 - Actually convert chosen start/end date filters to UTC datetime

bamcdole:
[DE44147](https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F602461243651&fdp=true?fdp=true)

We were converting the start/end dates to an ISO UTC datetime string (which _appeared_ to be correct, as it has the correct format and properly ended with a 'Z')

However, what we were missing was actually converting the start/end datetimes to UTC time.

For example, lets assume we chose **07/22/2021** as the start filter, and our time zone is **UTC+2:00** (this is what CIPD uses).

Before this fix, we would send the datetime `2021-07-22T00:00:00.000Z`, which is **WRONG**. We convert the completion date of records to UTC before sending them to the LMS for saving in the database, so any records saved on 07/22/2021 would have a completion date of `2021-07-21T22:00:00.000Z` (note the day is 21, not 22).

After this fix, we convert the filter datetime to UTC first, and instead send `2021-07-21T22:00:00.000Z` to our query API. This will correctly compare to the completion dates for records, and give us the expected results.


lweber-d2l (approved):
Looks good.
C/I failed a couple times for different reasons, 3rd time's a charm :-)

PR #318, opened 2021-07-22
DE44147 - Filter by Date showing records outside of selected range by one day

bamcdole:
Follow up PR for [DE44147](https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fdefect%2F602461243651&fdp=true?fdp=true) - I found another issue that produces this incorrect behaviour

It turns out that the UTCDateTime type in SQL does some unexpected rounding when you get down to hundredths of a second:

image

Guess what produces a datetime string with 999 milliseconds? 
Answer - The dateJs.endOf() method, which is used by the end date filter in CPD My Records:
https://search.d2l.dev/xref/Brightspace/continuous-professional-development/src/helpers/time-helper.js?r=b49d28d9#14

This fix omits milliseconds from our end filter, which is a level of granularity that we didn't need anyway. This should prevent datetime comparison errors between the _very end_ of one day, and the _very start_ of another.


lweber-d2l (approved):
Nice catch!  👍

PR #321, opened 2021-07-28
US130185 - Updated print options language, minor code cleanup

bamcdole:
[US130185](https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F604960682227&fdp=true?fdp=true)


lweber-d2l (approved):

src/components/cpd-my-records.js
```diff
@@ -389,41 +389,33 @@ class MyCpdRecords extends BaseMixin(LitElement) {
...
-				?opened=${this.showPrintOptionsDialog}
-				@d2l-dialog-close=${this.closePrintOptionsDialog}
-				title-text=${this.localize('myRecords:printRecordsDialog:title')}
-				width="400"
-			>
+			<d2l-dialog ?opened=${this.showPrintOptionsDialog}
+						@d2l-dialog-close=${this.closePrintOptionsDialog}
+						title-text=${this.localize('myRecords:printRecordsDialog:title')}
+						width="400">
+
```
jorge-ramirez-arredondo, line 396: Do we have a formatting standard that we're supposed to be following for front-end html code?

bamcdole: I'm not aware of one - I just found the "dangling tag ends" very offensive to my sensibilities...

jorge-ramirez-arredondo: We should probably agree on a standard, because I actually prefer the dangling tag ends, lol

PR #329, opened 2021-08-13
Add functionality to save as draft and filter drafts from summary

jorge-ramirez-arredondo:
[US128935](https://rally1.rallydev.com/#/431372673576d/custom/444667152236?detail=%2Fuserstory%2F601271804275&fdp=true?fdp=true)
[US128932](https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F601271800879&fdp=true?fdp=true)


src/components/cpd-record-dialog.js
```diff
@@ -355,8 +373,21 @@ class RecordDialog extends BaseMixin(LitElement) {
...
-				<d2l-button slot="footer" ?disabled="${this.saving || this.readOnly}" @click="${this.saveForm}" primary>${this.localize('shared:saveText')}</d2l-button>
-				<d2l-button slot="footer" @click="${this.cancelDialog}">${this.localize('shared:cancelText')}</d2l-button>
+				<div slot="footer" class="d2l-dialog-footer">
+					<span class="d2l-dialog-footer-left">
+						<d2l-button ?disabled="${this.saving || this.readOnly}" @click="${this.saveForm}" primary>${this.localize('shared:saveText')}</d2l-button>
+						<d2l-button @click="${this.cancelDialog}">${this.localize('shared:cancelText')}</d2l-button>
+					</span>
+					<span>
+						<d2l-button
+							primary
```
lweber-d2l, line 383: Should the Save Draft button be primary?  The Save button is primary, and I believe our convention is to have 1 primary button in a dialog.  The mockup on US128935 has just the main Save/Create Record button as primary.

src/helpers/recordStateEnum.js
```diff
@@ -0,0 +1,12 @@
+const recordStateEnum = Object.freeze({
+	Published: 1,
+	Draft: 2
+});
+
+const recordStateEnumParser = Object.freeze({
```
lweber-d2l, line 6: I don't see any other references to `recordStateEnumParser`.  Will it be referenced in a future change?

jorge-ramirez-arredondo: Hmmm, I added it just in case it's helpful for converting from the enum number to the name in the future, but I guess it might never be used.

We have a similar thing for tracking method [here](https://github.com/Brightspace/continuous-professional-development/blob/master/src/helpers/trackingMethodEnum.js#L6), but it looks like that hasn't been used yet. Do you think it would be better to remove it?

src/components/cpd-record-dialog.js
jorge-ramirez-arredondo: Oh yeah, looks like I just copied that from another button and forgot to remove it, lol.

src/helpers/recordStateEnum.js
lweber-d2l: Thanks for the clarification.  I don't feel strongly either way - ok to keep it there.  If we haven't used these parsers by the end of the feature/initiative, could remove them then.

lweber-d2l (approved):

jorge-ramirez-arredondo:
`v2.7.8`
Dependent on LMS: `20.21.10.32106`

PR #331, opened 2021-08-18
Add publish button and disable conditions

julianpettit:
Button appearance logic:
- If the record is being newly created or is published, use the save text
- If the record is a draft, use the publish text

- If the record is a being newly created or is a draft, show the save as draft button
- If the record is published, hide the save as draft button

https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F601271807045


src/components/cpd-record-dialog.js
```diff
@@ -375,16 +375,23 @@ class RecordDialog extends BaseMixin(LitElement) {
...
+						<d2l-button
+							?disabled="${this.saving || this.readOnly}"
+							@click="${this.saveForm}"
+							primary
+						>
+							${this.record && this.record.RecordStateId === recordStateEnum.Draft ? this.localize('shared:publishText') : this.localize('shared:saveText')}</d2l-button>
 						<d2l-button @click="${this.cancelDialog}">${this.localize('shared:cancelText')}</d2l-button>
 					</span>
 					<span>
+					${!this.record || this.record.RecordStateId === recordStateEnum.Draft ? html`
```
jorge-ramirez-arredondo, line 387: I think there is one situation you're missing here. When creating a new record based off of an award, `this.record` is set in order to give the user some default values to create their new record. https://github.com/Brightspace/continuous-professional-development/blob/master/src/components/cpd-record-dialog.js#L193

So, in this scenario, `this.record` is truthy, but there is no `this.record.RecordStateId` yet (because it's a new record).

jorge-ramirez-arredondo (approved):

PR #340, opened 2021-09-27
Improve CPD Main Page accessibility

nsalamad:
https://rally1.rallydev.com/#/431372673576d/iterationstatus?detail=%2Fuserstory%2F608873426509


package-lock.json
```diff
@@ -1147,9 +1147,9 @@
       }
     },
     "@brightspace-ui-labs/view-toggle": {
-      "version": "0.7.5",
```
jorge-ramirez-arredondo, line 1150: I think this is how it works. Even though it was written as `^0.7.4` before in the `package.json` file, npm still took the liberty of getting `0.7.5` on its own. So, you probably didn't have to update it to say `^0.7.6` after all. That being said, I don't think it hurts any to update that.

jorge-ramirez-arredondo (approved):

PR #367, opened 2021-12-07
US130935 - FRA to BSI

jorge-ramirez-arredondo:
[US130935](https://rally1.rallydev.com/#/?detail=/userstory/606865866871&fdp=true)

This PR updates CPD to work with a fully BSI setup:
* Update missing imports in some files that are now apparent
* Update some component properties to match the new backend changes
* Remove FRA and iframe specific logic
* Clean up no longer needed files
* Remove FRA logic in tests
* Remove build and release github workflows
* Remove unused packages and scripts

Should release simultaneously with https://github.com/Brightspace/lms/pull/16866


src/components/cpd-admin.js
```diff
@@ -1,5 +1,6 @@
 
 import '@brightspace-ui/core/components/tabs/tabs';
+import '@brightspace-ui/core/components/tabs/tab-panel';
```
bamcdole, line 3: How was `tab-panel` working before this? 😆

src/components/cpd-admin.js
```diff
@@ -10,15 +11,14 @@ import trackingMethodEnum from '../helpers/trackingMethodEnum';
...
-				*/
-
+			admin: {
+				type: Boolean
+			},
+			metadata: {
+				type: Boolean
+			},
+			trackingMethod: {
+				type: Number
```
bamcdole, line 21: I let out an audible sigh of relief when I saw this

bamcdole (approved):
Looks good - this was a great thing to take on for Inspiration 👍

src/components/cpd-admin.js
jorge-ramirez-arredondo: I think what was happening before was that we were loading the whole app whenever we went to any specific page. So even though some pages didn't actually specify all their imports properly, some of those components were still working because they were being loaded from a different page.

jorge-ramirez-arredondo:
Released in BSI version `20.22.1-147`

PR #378, opened 2022-01-07
DE46639 - Fix Job Title Targets page

jguann:
[DE46639](https://rally1.rallydev.com/#/431372673576d/search?detail=%2Fdefect%2F621645014665)

- All job titles had the same target hours/minutes because it was all redirecting to the current user's personal targets page
- Fixed to redirect to job title's targets

chrome_ySoDXyCZlg


lweber-d2l (approved):
Nice - tiny fix 😄

PR #382, opened 2022-01-18
DE45677 Record Dialog - Remove double scroll bar

jguann:
[DE45677 ](https://rally1.rallydev.com/#/12024503567d/search?detail=%2Fdefect%2F612260093783)

- Fix double scroll bar defect in Record Dialog
  - Caused by extra space accounted for date input even when the calendar was closed
  - Dialog will now extend scrollable area if necessary when calendar is open
- Add demo page for dialog

chrome_mx4mMPhaEl


src/services/demo-cpd-service.js
```diff
@@ -133,10 +133,18 @@ export class DemoCpdService {
...
+	}
+
 	static async getTargetRecords() {
 		return fetch('../../../../data/record_array.json', getOptions).then(r => r.json());
 	}
 
+	static getUserDraftPermissions() {
+		return Promise.resolve(true);
+	}
+
```
jguann, line 147: Adding demo functions to address warnings. This also renders the targets properly on `index.html` demo page.

jorge-ramirez-arredondo (approved):

PR #383, opened 2022-01-27
DE46992 - TinyMCE error when calling .clear() on htmleditor

bamcdole:
[DE46992](https://rally1.rallydev.com/#/?detail=/defect/623318817925&fdp=true)

Unfortunately, I haven't been able to reproduce this locally (BSI build issues). However, I think these changes are low risk enough that it is worth trying.

My thought process is that the HTML Editor might be "disposing" its reference to tinymce before the CPD dialog gets to call the `.clear()` method.

I would like to try this solution first, before merging our "back up plan", which is to prevent clearing the answer inputs when the dialog is closed.

I was also recommended to remove the `tinymce` static files line in our rollup file, and enforce a minimum version of 1.40 for `htmleditor`. Slack thread: https://d2l.slack.com/archives/C0PHG3QB0/p1643244047203200

**UPDATE**: I am reverting the version bumps - they started breaking our unit tests. Will speak with Grant Cleary about it on Monday.

If the async changes fail to resolve the problem, or if the rollup file changes bread CPD in production, I will spin up a new PR this afternoon to fix.


jguann (approved):

bamcdole:
Should be in BSI v20.22.3-68 or -69

PR #387, opened 2022-01-27
DE46992 - Conditionally render Questions when dialog is actually opened

bamcdole:
[DE46992](https://rally1.rallydev.com/#/?detail=/defect/623318817925&fdp=true)

The [last attempt](https://github.com/Brightspace/continuous-professional-development/pull/383) didn't work, so this is a new approach.

Since the [HTML Editor's `clear()` API](https://github.com/BrightspaceUI/htmleditor/#methods) is what triggers the error, I am now trying to avoid the need to call that method at all.

Instead, we now conditionally render the Questions section of the form, when the dialog is actually open. 

This causes the `htmleditor` instances to be removed on dialog close, and recreated when the dialog opens. Doing this forces the `htmleditor` instances to "refresh" with the correct starting value, instead of caching whatever was entered last.

**OLD Demo**:
itLxZ7HXgF

**NEW Demo**:
rNCredDDHn


src/components/cpd-record-dialog.js
```diff
@@ -359,10 +369,6 @@ class RecordDialog extends BaseMixin(LitElement) {
 		if (creditMinutes) creditMinutes.value = null;
 		const dateCompleted = await this.queryElementById(this.inputId.DateCompleted);
 		if (dateCompleted) dateCompleted.value = formatForDatePicker(new Date());
-		await Promise.all(this.questions.map(async question => {
-			const questionToClear = await this.queryElementById(`answerText_${question.Id}`);
-			if (questionToClear) questionToClear.clear();
```
bamcdole, line 364: This is the "problem line" - `questionToClear` is a single instance of `htmleditor`.

Something about this `.clear()` call during our dialog close is resulting in a null reference exception in the `htmleditor` code.

src/components/cpd-record-dialog.js
```diff
@@ -245,6 +249,11 @@ class RecordDialog extends BaseMixin(LitElement) {
 	}
 
 	async handleDialogClose(e) {
+		this._renderQuestions = false;
+		setTimeout(() => {
+			this._renderQuestions = true;
+		});
```
bamcdole, line 255: @jorge-ramirez-arredondo pointed out that the dialog won't calculate height correctly if they Questions aren't rendered at first.

Here we are forcing the `htmleditor`s to reload, once the dialog is closed. This produces the same behaviour, in less steps.

I also renamed the property to explicitly describe what it is controlling.

This approach is ugly, but works. The plan is to revert this change once a proper `htmleditor` fix has been implemented next week.

jguann (approved):

bamcdole:
BSI v20.22.3-71 (maybe -72)

PR #388, opened 2022-01-31
DE47045 - Stop event propagation when attachment dialog is closed

bamcdole:
[DE47045](https://rally1.rallydev.com/#/?detail=/defect/623824706481&fdp=true)

The issue was being caused by the `cpd-record-dialog` _also_ catching the `d2l-dialog-close` event that was being fired by the `cpd-attachments` dialog.

This PR both prevents the `cpd-attachments` close event from propagating to parent components, _and_ moves the "answer clearing" logic below a check that protects against events that didn't originate from the `cpd-record-dialog`.

The `.stopPropagation()` fix is the "correct" way to do this, but the other change also doesn't hurt.


jorge-ramirez-arredondo (approved):

lweber-d2l (approved):

bamcdole:
BSI v20.22.3-80

PR #389, opened 2022-02-01
US135815 Lit 2 Migration - Phase 1

jguann:
[US135815](https://rally1.rallydev.com/#/431372673576d/dashboard?detail=%2Fuserstory%2F623960400869&fdp=true)

This PR makes backwards-compatible changes to prepare this repo for a Lit 2 upgrade. Since the changes are backwards compatible, it can be merged right away.

Non-breaking change checklist:
- [x] Add missing `.js` extensions for lit-html or lit-element imports
- [x] `await requestUpdate` -> `await elem.updateComplete`
  - No usages. 
- [x] add non-null checks to shadowRoot accesses (not including test files)
- [x] audit usages of `unsafeHTML` to ensure `null`, `undefined` inputs are handled as expected
  - No changes necessary (used in `cpd-record-report.js` and `cpd-report-logo.js`)
- [x] remove dynamic code from inside  `<template>` tags (Lit files only)
  - No usages. 
- [x] remove any dependencies on old Lit versions
  - All accounted for in migration:
   image
- [x] audit `ifDefined` usages for null parameters
  - No changes necessary, want same behaviour for null parameters (used in `cpd.js` and `cpd-my-records.js`)

For more information on the Lit 2 upgrade, visit https://lit.dev/docs/releases/upgrade/ or contact team Gaudi.


src/components/cpd-admin-items.js
```diff
@@ -263,7 +265,7 @@ class CpdAdminItems extends BaseMixin(LitElement) {
 	}
 
 	_getUpButton(index) {
-		const upButtons = this.shadowRoot.querySelectorAll('#up-button');
+		const upButtons = this.shadowRoot && this.shadowRoot.querySelectorAll('#up-button');
 		for (let i = 0; i < upButtons.length; i++) {
```
jorge-ramirez-arredondo, line 269: This whole method will still fail if `this.shadowRoot` is `undefined` since it's dependent on the result. In order for this check to be more helpful, we would have to change the method to return `null` in the case the `this.shadowRoot` is `undefined` and we would also have to update any methods calling on this one to accept a `null` return from here.

I think in this case, we shouldn't worry about whether `this.shadowRoot` is `undefined`. This method only seems to be called from a button press, so `this.shadowRoot` should be defined well before then.

jorge-ramirez-arredondo: Hmmm, now that I've looked through the rest of the PR more though, there seem to be a lot of places where this is the case. Considering that, I would say that it's fine to just go with the change you made here. Having to go through each scenario to identify if the `undefined` check will actually be helpful or not is probably not worth the effort.

jguann: How about only adding `if (!this.shadowRoot) return;` to the beginning of the method instead?

lweber-d2l (commented):
These changes all look good so far.

src/components/cpd-admin-items.js
jorge-ramirez-arredondo: Just to recap our zoom conversation:
We can leave the changes the way you have them now, but maybe just add a comment (PR comment, not code comment) to the constructor of each file with `this.shadowRoot` calls that explains whether there is any `this.shadowRoot` calls in the constructor or not.

src/components/cpd-admin-items.js
```diff
@@ -115,7 +115,9 @@ class CpdAdminItems extends BaseMixin(LitElement) {
 		this.objectData = e.target.getAttribute('item-json') &&
 			JSON.parse(e.target.getAttribute('item-json')) || {};
 		this.inputField = this.objectData[this.context.textFieldName] || '';
-		this.shadowRoot.querySelector(`#create-${this.context.type}-dialog`).open();
```
jguann, line 118: There are no calls to `this.shadowRoot` from `constructor()` above.

src/components/page-select.js
```diff
@@ -78,14 +78,14 @@ class PageSelect extends BaseMixin(LitElement) {
```
jguann, line 78: There are no calls to `this.shadowRoot` from `constructor()` above.

src/components/cpd.js
```diff
@@ -113,7 +113,7 @@ class Cpd extends BaseMixin(LitElement) {
 	}
```
jguann, line 113: There are no calls to `this.shadowRoot` from `constructor()` above.

src/components/cpd-record-dialog.js
```diff
@@ -324,7 +324,7 @@ class RecordDialog extends BaseMixin(LitElement) {
 	}
```
jguann, line 324: There are no calls to `this.shadowRoot` in `constructor()` above.

src/components/attachments.js
```diff
@@ -78,7 +78,11 @@ class Attachments extends BaseMixin(LitElement) {
 		const newFileArray = [...this.currentAttachments];
```
jguann, line 78: There are no calls to `this.shadowRoot` from `constructor()` above.

src/components/cpd-manage-targets.js
```diff
@@ -170,7 +170,9 @@ class ManageCpdTargets extends BaseMixin(LitElement) {
 					|| this.invalidUnstructuredRecords;
```
jguann, line 170: There are no calls to `this.shadowRoot` from `constructor()` above.

jguann (commented):
Noting that `this.shadowRoot` calls do not occur before `connectedCallback`.

jorge-ramirez-arredondo (approved):

